PR #7, opened 2021-11-07
Randomize ports. Fixes #3

### LINKED ISSUE #3: Make sure ports are randomized
Otherwise multiple lemmy's can't run on the same machine.

dessalines:
I tested this and it works.

### DIFF
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -61,6 +61,11 @@
         owner: '991'
 
   - block:
+    - set_fact:
+        lemmy_port: "{{ 32767 |random(start=1024) }}"
+        lemmy_ui_port: "{{ 32767 |random(start=1024) }}"
+        pictrs_port: "{{ 32767 |random(start=1024) }}"
+
     - name: add template files
       template:
         src: '{{item.src}}'
@@ -76,8 +81,6 @@
       vars:
         lemmy_docker_image: "dessalines/lemmy:{{ lemmy_version | default( lookup('file', 'VERSION') )}}"
         lemmy_docker_ui_image: "dessalines/lemmy-ui:{{ lemmy_ui_version | default(lemmy_version | default(lookup('file', 'VERSION')))}}"
-        lemmy_port: "8536"
-        lemmy_ui_port: "1235"
 
     - name: add the config.hjson
       template:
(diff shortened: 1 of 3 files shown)


templates/nginx.conf
```diff
@@ -27,7 +27,7 @@ server {
     ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256';
     ssl_session_timeout  10m;
     ssl_session_cache shared:SSL:10m;
-    ssl_session_tickets off;
+    ssl_session_tickets on;
```
dessalines, line 30: I was getting an SSL error very strangely on my test host, and this was the fix.

lemmy.yml
```diff
@@ -61,6 +61,11 @@
         owner: '991'
 
   - block:
+    - set_fact:
+        lemmy_port: "{{ 32767 |random(start=1024) }}"
+        lemmy_ui_port: "{{ 32767 |random(start=1024) }}"
+        pictrs_port: "{{ 32767 |random(start=1024) }}"
```
Nutomic, line 67: This will generate a new, random port on each run. I believe this could cause errors if docker-compose already uses the new port, but nginx still tries to connect to the old one. Plus it will cause unnecessary changes in diffs. Besides, if the generated port is already taken, it will cause errors.

So i think it would be best to generate the ports only on first run, and then store them in a file for subsequent runs, similar to the [password lookup](https://github.com/LemmyNet/lemmy-ansible/blob/main/lemmy.yml#L90). Unfortunately i cant find an easy way to do that.

dessalines: The nginx is also overwritten and reloaded, so that should't be a problem.

Nutomic: That is a separate step, so there will be some time between docker port changing, and nginx port changing. On a big instance that will definitely affect some users. And it will probably also break websocket connections.

dessalines: It pry wouldn't take more than 20 seconds. I'll test websockets right now.

Nutomic:
I dont think this is ideal but okay.

PR #14, opened 2021-11-09
Route any of the application accepts to lemmy.

dessalines:


Nutomic:
Great, this is probably even better than my proposal.

ansible.cfg
```diff
@@ -1,6 +1,6 @@
 [defaults]
 inventory = inventory
-# interpreter_python = /usr/bin/python3
+interpreter_python = /usr/bin/python
```
Nutomic, line 3: Should this really be committed?

templates/nginx.conf
```diff
@@ -60,10 +60,7 @@ server {
       # lemmy_port: 8536
 
       set $proxpass "http://0.0.0.0:{{lemmy_ui_port}}";
-      if ($http_accept = "application/activity+json") {
-        set $proxpass "http://0.0.0.0:{{lemmy_port}}";
-      }
-      if ($http_accept = "application/ld+json; profile=\"https://www.w3.org/ns/activitystreams\"") {
+      if ($http_accept ~ "(application\/activity\+json|application\/ld\+json)") {
```
Nutomic, line 63: `if ($http_accept ~ "application/.*") ` works fine, please use that (already active on ds9.lemmy.ml)

Nutomic:
Great, this is much better than my proposal.

Issue #15, opened 2021-11-23
Error: sudo: password required


calbasi:
Ansible needs a sudo password to avoid this error: 

> TASK [install python for Ansible] ******************************************************************************************
> fatal: [my_user@my_server]: FAILED! => {"changed": true, "msg": "non-zero return code", "rc": 1, "stderr": "Shared connection to my_server closed.\r\n", "stderr_lines": ["Shared connection to my_server closed."], "stdout": "sudo: es requereix una contrasenya\r\r\n", "stdout_lines": ["sudo: es requereix una contrasenya", ""]}

It can be managed with -K parameter:

`ansible-playbook -i inventory/hosts lemmy.yml --become -K`

PR #16, opened 2021-11-23
Error: sudo: password required

calbasi:
This PR addresses issue #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ To run this ansible playbook, you need to:
 - Have a server / VPS where lemmy will run.
 - Configure a DNS `A` Record to point at your server's IP address.
 - Make sure you can ssh to it, with a sudo user: `ssh <your-user>@<your-domain>`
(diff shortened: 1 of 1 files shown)


dessalines:
Seems good to me, although many setups will just use the root user so won't need this. Thanks!

Nutomic:
I would prefer to specify sudo/become options in ansible.cfg, to keep the command which users have to copy shorter.

Should be something like `become: True` and `become_ask_pass: True`.

calbasi:
I understand your point of view. But we need to do something to avoid headaches to users who usually don't work with Ansible.

Nutomic:
The behavior will be identical whether you set the params via command line (like you did in this PR), or via ansible.cfg. But the latter seems simpler for new users, as they dont have to copy extra parameters.

dessalines:
@Nutomic if you do a pr for that we can merge.

PR #17, opened 2021-11-25
Move privilege escalation params into config

Nutomic:
@calbasi Can you confirm that this works for you with `ansible-playbook -i inventory/hosts lemmy.yml`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Edit the inventory hosts file (inventory/hosts) to your liking.
 
 Run the playbook: 
 
-`ansible-playbook -i inventory/hosts lemmy.yml --become -K`
+`ansible-playbook -i inventory/hosts lemmy.yml`
 
 If the command above fails, you may need to comment out this line In the ansible.cfg file:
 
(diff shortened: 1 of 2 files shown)


calbasi:
> 

I think so. I have been asked for my password. Here you are the output:

> $ ansible-playbook -i inventory/hosts lemmy.yml
> BECOME password: 
> [WARNING]: Unable to parse
> /media/magatzem/magatzem_joan/tmp/kk/lemmy_patched/lemmy-ansible-config-
> params/inventory/hosts as an inventory source
> [WARNING]: No inventory was parsed, only implicit localhost is available
> [WARNING]: provided hosts list is empty, only localhost is available. Note that
> the implicit localhost does not match 'all'
> 
> PLAY [all] *********************************************************************
> skipping: no hosts matched
> 
> PLAY RECAP *********************************************************************
>

Nutomic:
Good, but as the warnings say, there is something wrong with your inventory file so it wont work.

Nutomic:
Actually this is not so great, it asks for root password even when none is set on the server. So this is very annoying for everyone who connects directly as root, or uses passwordless sudo. I think it would be better to set only `become = True`, and mention `-K` param somewhere in the docs for those who need it.

dessalines:
Ok, I'll do a PR for that.

PR #18, opened 2021-11-25
Adding a note about priviledges to README.

dessalines:
cc @calbasi

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Run the playbook:
 
 *Note*: if you are not the root user or don't have password-less sudo, use this command:
 
-`ansible-playbook -i inventory/hosts lemmy.yml --become -K`
+`ansible-playbook -i inventory/hosts lemmy.yml --ask-become-pass`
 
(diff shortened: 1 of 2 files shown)


ansible.cfg
```diff
@@ -4,7 +4,3 @@ inventory = inventory
 
 [ssh_connection]
 pipelining = True
-
-[privilege_escalation]
-become = True
```
Nutomic, line 9: You can leave the `become = True`, and remove `--become` above. This one doesnt cause any prompt, so there is no disadvantage setting it.

README.md
```diff
@@ -40,6 +40,10 @@ Run the playbook:
 
 `ansible-playbook -i inventory/hosts lemmy.yml`
 
+*Note*: if you are not the root user or don't have password-less sudo, use this command:
+
+`ansible-playbook -i inventory/hosts lemmy.yml --become -K`
```
Nutomic, line 45: Could also use `--ask-become-pass` instead of `-K`, seems clearer.

PR #29, opened 2022-03-27
Rename rate_limit with domain name space

fadelkon:
This allows to provision one lemmy instance in a server where there are already other lemmy instances installed with this same nginx template.
Currently, nginx refuses to reload starting at the 2d instance installed, because lemmy_ratelimit is already defined in another site conf.
It also makes sense to just give each lemmy its own rate limit strategy instead of sharing a host-global one.


Nutomic:
Thank you!

PR #30, opened 2022-03-27
Render nginx templates into sites-available

fadelkon:
Render nginx templates into sites-available and then symlink them.
Instead of rendering them directly to sites-enabled.

It's a good practice and makes it easier for disabling or substituting temporarily a site config.

Currently, the sites-enabled file is a regular file. If we just try to replace it with a symlink, it fails. To avoid upgrade conflicts, i add 2 tasks to remove the file if it's regular instead of a symlink. We could just delete it always before symlinking, but then we would generate unnecessary activity and "changed" noise in the output.

Tested with a running instance.

### DIFF
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -65,12 +65,30 @@
           dest: '{{lemmy_base_dir}}/{{domain}}/docker-compose.yml'
           mode: '0600'
         - src: 'templates/nginx.conf'
-          dest: '/etc/nginx/sites-enabled/{{domain}}.conf'
+          dest: '/etc/nginx/sites-available/{{domain}}.conf'
           mode: '0644'
       vars:
         lemmy_docker_image: "dessalines/lemmy:{{ lemmy_version | default( lookup('file', 'VERSION') )}}"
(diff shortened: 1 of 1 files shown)


Nutomic:
To be honest, i have never really used the sites-available folder, nor did i need it. For disabling a site, it works just as well to move the config from sites-enabled to sites-enabled, and later move it back to enable. That said, your implementation looks good and i'm not really opposed to merging it. But i also dont really see the benefit.

fadelkon:
The benefit for me is the order and coherence with other sites I (may) have in the same nginx. I always do the enabled -> available thing, and I'm used to it.

It's also a convention at least at Debian and Ubuntu, as the nginx package creates these folders and incentives this. So a greater benefit would be... Compliance with conventions?

In the other hand, it's not that important and you may prefer to prioritise brevity, that has its own benefits.

At the end, that's why I prefer doing this by hand better than scripted by somebody else that may have different priorities :)


El 28 de març de 2022 16:29:13 CEST, Nutomic ***@***.***> ha escrit:

dessalines:
I spose it is the standard convention, and I'm good with it.

Issue #35, opened 2022-04-09
"The conditional check 'reg_enabled['stat isreg']' failed. The error was: error while evaluating conditional (reg_enabled['stat isreg']): 'dict object' has no attribute 'stat isreg'


DMTryptamines:
Was setting up a new instance and ran into this after running the ansible playbook.

```
fatal: [root@ip.ip.ip.ip]: FAILED! => {
    "msg": "The conditional check 'reg_enabled['stat isreg']' failed. The error was: error while evaluating conditional (reg_enabled['stat isreg']): 'dict object' has no attribute 'stat isreg'\n\nThe error appears to be in '/home/ubuntu/lemmy-ansible/lemmy.yml': line 80, column 9, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n      - name: remove if regular file (legacy) instead of symlink\n        ^ here\n"
}
```

The issue is exactly where it says it is and changing `when: reg_enabled.stat.isreg` to `when: reg_enabled` in `lemmy.yml` fixes the problem and let's the install finish. Doesn't appear to be any issues with changing......so far.

PR #36, opened 2022-04-10
Update lemmy.yml

DMTryptamines:
https://github.com/LemmyNet/lemmy-ansible/issues/35

### DIFF
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -81,7 +81,7 @@
         file:
           path: "/etc/nginx/sites-enabled/{{domain}}.conf"
           state: absent
-        when: reg_enabled.stat.isreg
+        when: reg_enabled
 
       - name: enable nginx site
         file:
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -81,7 +81,7 @@
         file:
           path: "/etc/nginx/sites-enabled/{{domain}}.conf"
           state: absent
-        when: reg_enabled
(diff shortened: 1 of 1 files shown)


Nutomic:
This doesnt look correct. I think the problem is that `isreg` is only set if the file exists. So the correct condition is probably `reg_enabled.stat.exists and reg_enabled.stat.isreg`.

cc @fadelkon

fadelkon:
El 11 d’abril de 2022 14:10:40 CEST, Nutomic ***@***.***> ha escrit:
>This doesnt look correct. I think the problem is that `isreg` is only set if the file exists.

I see

>  So the correct condition is probably `reg_enabled.stat.exists and reg_enabled.stat.isreg`.
>
>cc @fadelkon 
>

Sure, makes sense!

lemmy.yml
```diff
@@ -81,7 +81,7 @@
         file:
           path: "/etc/nginx/sites-enabled/{{domain}}.conf"
           state: absent
-        when: reg_enabled.stat.isreg
```
dessalines, line 84: reg_enabled.stat.exists and reg_enabled.stat.isreg

dessalines (changes requested):

DMTryptamines (commented):
Tested and can confirm r`eg_enabled.stat.exists and reg_enabled.stat.isreg` completed a fresh install fine where 'reg_enabled.stat.isreg` alone did not.

dessalines:
Thanks!

PR #49, opened 2022-07-14
Add logging limitation

pixlguru:
Prevents growing of docker/overlay2 due to logfiles piling up

### DIFF
--- a/templates/docker-compose.yml
+++ b/templates/docker-compose.yml
@@ -13,6 +13,10 @@ services:
     depends_on:
       - postgres
       - pictrs
+    logging:
+      options:
+        max-size: "100m"
+        max-file: "10"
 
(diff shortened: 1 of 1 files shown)


Nutomic:
Makes sense. But i think its much easier to configure this via `/etc/docker/daemon.json` file.

https://docs.docker.com/config/containers/logging/local/

templates/docker-compose.yml
```diff
@@ -13,6 +13,10 @@ services:
     depends_on:
       - postgres
       - pictrs
+    logging:
+      options:
```
dessalines, line 17: 20m, 5 files.

dessalines (changes requested):

dessalines (approved):

PR #59, opened 2023-01-26
V0.17.0 changes

dessalines:
Okay I just tested `0.17.0-rc.1` on a test server with this PR, and it works great.

On the release, the `VERSION` will get changed, and this will have to be merged.


lemmy.yml
```diff
@@ -21,15 +21,54 @@
...
+  - name: Install Docker Module for Python
+    pip:
+      name: docker
+
+  - name: Install or upgrade docker-compose
+    get_url: 
+      url : "https://github.com/docker/compose/releases/download/1.25.5/docker-compose-Linux-x86_64"
+      dest: /usr/local/bin/docker-compose
+      mode: 'a+x'
+      force: yes
```
Nutomic, line 71: Why are you switching the Docker install method? Afaik there is no problem with Docker from Ubuntu repos and it has the major advantage that its very easy to install and upgrade.

templates/docker-compose.yml
```diff
@@ -20,20 +20,44 @@ services:
...
+      "-c", "default_statistics_target=100",
+      "-c", "random_page_cost=4",
+      "-c", "effective_io_concurrency=2",
+      "-c", "work_mem=4GB",
+      "-c", "min_wal_size=1GB",
+      "-c", "max_wal_size=30GB",
+      "-c", "max_worker_processes=4",
+      "-c", "max_parallel_workers_per_gather=2",
+      "-c", "max_parallel_workers=4",
+      "-c", "max_parallel_maintenance_workers=2"
```
Nutomic, line 59: Are these settings a good default for all servers? Otherwise it would be better to define it in a different file which can be edited by the user without creating git conflicts.

lemmy.yml
dessalines: I tried on a brand new test server, and it completely fails now. Using the docker repos on ubuntu or debian is never the recommended way anymore, and it asks you to install using their official method. This was taken from an ansible example that uses that official install method.

templates/docker-compose.yml
dessalines: I'll remove... doing this as an external postgres settings file is a better way to do it anyway.

Nutomic (approved):

lemmy.yml
Nutomic: Hmm alright.

dessalines:
Oops, mis-clicked and accidentally merged. I reverted this tho. We just have to remember to re-merge it when we're ready for the release.

Issue #67, opened 2023-05-05
TASK [start docker-compose] : Unable to load docker-compose


tifriis208:
Hello,

On a fresh lemmy install on ubuntu server 22.04 (arm64 for raspberry pi and x64), I got this error (each time):
```
TASK [start docker-compose] ************************************************************************************************************************************************************
fatal: [user@IP_ADDRESS]: FAILED! => {"changed": false, "msg": "Unable to load docker-compose. Try `pip install docker-compose`. Error: Traceback (most recent call last):\n  File \"/tmp/ansible_docker_compose_payload_ixgojrhn/ansible_docker_compose_payload.zip/ansible_collections/community/docker/plugins/modules/docker_compose.py\", line 506, in <module>\nModuleNotFoundError: No module named 'compose'\n"}
```

For information, I also got this warning:
```
 TASK [Add Docker GPG apt Key] ******************************************************************************************************************
[WARNING]: Module remote_tmp /root/.ansible/tmp did not exist and was created with a mode of 0700, this may cause issues when running as
another user. To avoid this, create the remote_tmp dir with the correct permissions manually
changed: [user@IP_ADDRESS]
```

Any idea how to run this playbook without error ?
Thanks :)

tifriis208:
I try to install `ansible-galaxy collection install community.docker` but nothing change

dessalines:
I've never tested this playbook on arm, so you might have to install docker manually on that machine, then skip the docker install steps in the playbook.

tifriis208:
Thanks for your return. Indeed, the playbook only work for x86_64: 
  ```
- name: Install or upgrade docker-compose
    get_url: 
      url : "https://github.com/docker/compose/releases/download/1.25.5/docker-compose-Linux-x86_64"
```

As raspberry pi is officialy supported in the documentation (`Supports arm64 / Raspberry Pi`), would you accept a pull request to modify this playbook according to [official docker documentation](https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository)?

Thanks

tifriis208:
I think this Pull Request should also solved this issue : https://github.com/LemmyNet/lemmy-ansible/pull/62

dessalines:
Sure. 

#62 didn't work when I used a fresh ubuntu install, and no docker installation recommends you using the docker from the ubuntu repos.

jonaharagon:
[...]

PR #69, opened 2023-06-02
Install latest docker/docker-compose via pip. Proposed as fix for #67

RedKrieg:
I was getting the same docker-compose error seen in #67 .  This installs `docker-compose` the same way that `docker` is installed, and a proper entry point is already created in the expected location `/usr/local/bin/docker-compose`.  I have been able to complete and test a deployment on a fresh ubuntu-22.04-x86_64 instance with this change.

### DIFF
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -62,13 +62,12 @@
   - name: Install Docker Module for Python
     pip:
       name: docker
+      state: latest
 
   - name: Install or upgrade docker-compose
-    get_url: 
-      url : "https://github.com/docker/compose/releases/download/1.25.5/docker-compose-Linux-x86_64"
-      dest: /usr/local/bin/docker-compose
-      mode: 'a+x'
-      force: yes
+    pip:
+      name: docker-compose
+      state: latest
 
   - name: copy docker config
     copy: src='../files/docker-daemon.json' dest='/etc/docker/daemon.json' mode='0644'
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -59,14 +59,11 @@
       state: latest
       update_cache: true
 
-  - name: Install Docker Module for Python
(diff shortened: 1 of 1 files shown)


lemmy.yml
```diff
@@ -62,13 +62,12 @@
...
 
   - name: Install or upgrade docker-compose
-    get_url: 
-      url : "https://github.com/docker/compose/releases/download/1.25.5/docker-compose-Linux-x86_64"
-      dest: /usr/local/bin/docker-compose
-      mode: 'a+x'
-      force: yes
+    pip:
+      name: docker-compose
+      state: latest
```
Nutomic, line 70: You should be able to install multiple packages in a single command like this:
```
- name: Install multi python packages with version specifiers
  ansible.builtin.pip:
    name:
      - django>1.11.0,<1.12.0
      - bottle>0.10,<0.20,!=0.11
```

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html

RedKrieg: I kept them in their own tasks purely to avoid altering the task structure.  I've pushed an updated commit to combine these in to a single task.  Thanks for the review.

RedKrieg:
I know that the pip installed version of docker-compose is stuck at `1.x` while the binary that this patch replaces is `2.x`, but ansible's `docker_compose` module actually requires `docker-compose>= 1.7.0,< 2.0.0` per [their documentation](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_compose_module.html#requirements).  Using the binary with this module is unsupported (and the cause of #67 ).

dessalines (approved):
I'm good with this, I really do hope it works with existing docker installations tho.

jonaharagon (approved):
Running this script on my existing installation worked fine 👍

PR #71, opened 2023-06-05
Add inbox to nginx.conf backend routes

pineapplemachine:
I can see in the nginx log that requests are being sent to an `/inbox` route, and it looks like these requests ought to be directed to the backend. This PR makes a small change to nginx.conf to make this happen by default.

See https://lemmy.ml/post/1148327?scrollToComments=true

### DIFF
--- a/templates/nginx.conf
+++ b/templates/nginx.conf
@@ -77,7 +77,7 @@ server {
     }
 
     # backend
-    location ~ ^/(api|pictrs|feeds|nodeinfo|.well-known) {
+    location ~ ^/(api|pictrs|feeds|inbox|nodeinfo|.well-known) {
       proxy_pass http://0.0.0.0:{{lemmy_port}};
       proxy_http_version 1.1;
       proxy_set_header Upgrade $http_upgrade;


templates/nginx.conf
```diff
@@ -77,7 +77,7 @@ server {
     }
 
     # backend
-    location ~ ^/(api|pictrs|feeds|nodeinfo|.well-known) {
+    location ~ ^/(api|pictrs|feeds|inbox|nodeinfo|.well-known) {
```
dessalines, line 80: @Nutomic Is this one necessary? Wouldn't federation be not working anywhere if this was required?

pineapplemachine: I'm not really sure if this change helps or hurts, but I can at least confirm that federation is not working the way I'd expect it to on the instance I set up recently using this repo. The main issue is that comment syncing, in that most comments that are visible on other instances are not visible on my own instance when I visit a post.

There's more info here about comment sync problems that I and others have been experiencing: https://lemmy.ml/post/1148327?scrollToComments=true

pineapplemachine: Here's an example of traffic logged by nginx when I visit `mylemmydomain.com/inbox`:

```
[my ip address] - - [05/Jun/2023:15:51:14 +0000] "GET /inbox HTTP/2.0" 404 0 "-" "Mozilla/5.0 [my user-agent]..."
[my ip address] - - [05/Jun/2023:15:51:14 +0000] "GET /inbox HTTP/2.0" 404 0 "-" "Mozilla/5.0 [my user-agent]..."
```

Here's an example of traffic that occurs regularly, with a variety of IPs and user-agent strings, most of which include either "lemmy.ml" or "mastodon.social", that made me think this traffic was probably not meant for the frontend:

```
[a datacenter ip address] - - [05/Jun/2023:15:50:42 +0000] "POST /inbox HTTP/1.1" 400 0 "-" "http.rb/5.1.1 (Mastodon/4.1.2+nightly-20230523; +https://mastodon.social/)"
[a datacenter ip address] - - [05/Jun/2023:15:50:42 +0000] "POST /inbox HTTP/1.1" 400 0 "-" "http.rb/5.1.1 (Mastodon/4.1.2+nightly-20230523; +https://mastodon.social/)"
```

pineapplemachine: Sorry, one more comment: I can see now that the comment syncing behavior is probably related to this. #2825

But I am still wondering why I'm seeing so much traffic on the `/inbox` route for an instance where I am currently the only registered user.

Nutomic: Yep this is already handled by `$request_method = POST` above. The inbox is where incoming federation actions from other servers are received (as POST requests).

dessalines (changes requested):
Changing this to request changes, because it breaks the front end inbox route.

Nutomic:
Yes this can be closed.

PR #73, opened 2023-06-05
 Increase pictrs mem_limit

pixlguru:
Fixes pictrs crashing / OOM

### DIFF
--- a/templates/docker-compose.yml
+++ b/templates/docker-compose.yml
@@ -47,7 +47,7 @@ services:
     volumes:
       - ./volumes/pictrs:/mnt
     restart: always
-    mem_limit: 200m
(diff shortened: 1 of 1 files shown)


Nutomic:
@asonix Can you say whats a good default value for this?

pixlguru:
Pictrs has a mem usage averaged out at about 400MiB, I guess we need some headroom

asonix:
When idling pict-rs doesn't need much ram, but if it calls into imagemagick or ffmpeg ram use spikes for the duration. I think I give mine around 1G, but lower values might be okay

dessalines (approved):
We can always increase it later if necessary. Thanks for this.

PR #75, opened 2023-06-06
Cache responses for unauthenticated users

Nutomic:
This can potentially reduce server load a lot because unauthenticated users trigger much fewer database queries. It will only be effective once lemmy-ui switches from websocket and uses the http api (Lemmy 0.18). 

It works fine in a quick test, but will need more testing to make sure it doesnt accidentally leak private data. For now I deployed it on ds9.lemmy.ml


dessalines (commented):
I wonder what the best way to test this is, to make sure nginx is properly doing caching. 

Our isomorphic server, doing local requests, makes this a little complicated to think about.

miyurusankalpa:
maybe caching static content like images for all users with a larger valid period will also help reduce the load.

Nutomic:
@dessalines This PR sets the header `x-cache-status` which you can check. See the nginx docs for `$upstream_cache_status` for details.

templates/nginx.conf
```diff
@@ -90,6 +98,13 @@ server {
       proxy_set_header X-Real-IP $remote_addr;
       proxy_set_header Host $host;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+
+      # Response cache for unauthenticated users
+      proxy_cache lemmy_cache_{{domain}};
+      proxy_cache_use_stale error timeout http_500 http_502 http_503 http_504;
```
ticoombs, line 104: adding `updating` can also help here to reduce requests to upstream, when an object expires out of the cache.

templates/nginx.conf
```diff
@@ -74,6 +75,13 @@ server {
       proxy_set_header X-Real-IP $remote_addr;
       proxy_set_header Host $host;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+
+      # Response cache for unauthenticated users
+      proxy_cache lemmy_cache_{{domain}};
+      proxy_cache_use_stale error timeout updating http_500 http_502 http_503 http_504;
+      proxy_cache_valid 1m;
```
dessalines, line 82: I'd be up for increasing this also.

@Nutomic anything preventing merging this?

Nutomic: I deployed it on https://voyager.lemmy.ml/ just now, and it seems to work well. Data is getting cached as expected when user is logged out, and not cached when user is logged in. I also didnt find any leakage of private data.

Though it would be good to update it for using https://github.com/LemmyNet/lemmy-ui/pull/1641 once that is deployed as its a bit cleaner.

Nutomic:
I had this deployed on lemmy.ml but its not working. The site was logging me out whenever I opened a new tab, and sometimes there were generic error messages which wouldnt go away. We have to wait until there is proper caching support in the projects themselves.

PR #78, opened 2023-06-06
Cache static content

miyurusankalpa:
Fixes LemmyNet/lemmy-ui#1098

I started the PR using some configs I used for my own projects, but need to setup a lemmy test environment to test the changes.

Also feel free to suggest changes to cache days.

The cache control headers is best to set from the lemmy-ui it self, instead of changing them here.

### DIFF
--- a/templates/nginx.conf
+++ b/templates/nginx.conf
@@ -1,4 +1,5 @@
 limit_req_zone $binary_remote_addr zone={{domain}}_ratelimit:10m rate=1r/s;
+proxy_cache_path /var/cache/lemmy/static/{{domain}} levels=1:2 keys_zone=lemmy_static_cache_{{domain}}:10m max_size=2g inactive=2d use_temp_path=off;
 
 server {
     listen 80;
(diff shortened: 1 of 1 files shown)


templates/nginx.conf
```diff
@@ -70,6 +71,22 @@ server {
...
+        proxy_cache_revalidate on;
+        proxy_cache_background_update on;
+        proxy_cache_lock on;
+
+        proxy_cache_valid 200 1d; #only cache 200s
+
+        proxy_ignore_headers Cache-Control;
+        add_header Cache-Control "public, max-age=86400, stale-while-revalidate=10" always;  #cache for 1 day
+        add_header X-Cache-Status $upstream_cache_status;
+      }
```
Nutomic, line 88: Its better to fix these things in the respective repos so that they send proper cache-control headers. Putting it here makes things unnecessarily complicated.

templates/nginx.conf
```diff
@@ -86,6 +101,20 @@ server {
...
+        proxy_cache_use_stale error updating timeout http_500 http_502 http_503 http_504;
+
+        proxy_cache_revalidate on;
+        proxy_cache_background_update on;
+        proxy_cache_lock on;
+
+        proxy_cache_valid 200 1d; #only cache 200s
+
+        add_header X-Cache-Status $upstream_cache_status;
+      }
```
Nutomic, line 116: Here its the same, pictrs should send the right cache headers. I thought it already does, maybe we need to use a newer version. Here is the project repo: https://git.asonix.dog/asonix/pict-rs/

Nutomic: Also the images and assets are not causing much server load so Im not sure its worth the trouble caching them. My PR #75 caches API responses and HTML so it will reduce server load a lot as most users can be served from the cache instead of causing SQL queries.

miyurusankalpa: agreed. 
I did a test with pict-rs and it looks like it does some caching on it own as well. no need to have a cache on nginx side as well.

Nutomic: Great, gonna close this pr then.

PR #88, opened 2023-06-19
Added support for Red Hat-based systems in lemmy.yml

christopher-conley:
I wanted to self-host a Lemmy instance, but I saw that the Ansible playbook for it only supports Debian-based systems, so I expanded out the playbook to support Red Hat-based distributions. I've tested it on two different VPS providers (Vultr & Digital Ocean) on both existing and brand-new from-scratch instances. It runs and deploys successfully, and I have a Lemmy instance running now that was deployed via this playbook and is federating with other instances.

### DIFF
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -3,16 +3,25 @@
 
   # Install python if required
   # https://www.josharcher.uk/code/ansible-python-connection-failure-ubuntu-server-1604/
-  gather_facts: False
+  gather_facts: True
   pre_tasks:
     - name: check lemmy_base_dir
       fail:
         msg: "`lemmy_base_dir` is unset. if you are upgrading from an older version, add `lemmy_base_dir=/lemmy` to your inventory file."
       when: lemmy_base_dir is not defined
 
-    - name: install python for Ansible
+    - name: install python for Ansible (Debian)
+      when: ansible_os_family == "Debian"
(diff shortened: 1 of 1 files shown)


Nutomic:
Im sceptical about this because dessalines and I have zero experience with Red Hat, so we wouldnt be able to maintain this. Fact is that the playbook is targeted at inexperienced users who setup a VPS specifically for Lemmy, so they can select Ubuntu during installation. If you know what you are doing, use the manual install.

christopher-conley:
Also just amended this PR to fix Issue #89

Nutomic:
I would consider this if the changes were minimal, but as is the size of the playbook is almost doubled. This means every change needs to be adjusted and tested for two completely different types of distros, which doubles the maintenance burden. Not worth the trouble imo. You can put a Redhat playbook in your own repo, and link that from the Lemmy docs instead.

christopher-conley:
> I would consider this if the changes were minimal, but as is the size of the playbook is almost doubled. This means every change needs to be adjusted and tested for two completely different types of distros, which doubles the maintenance burden. Not worth the trouble imo. You can put a Redhat playbook in your own repo, and link that from the Lemmy docs instead.

Ok. Lemmy docs being the README in the lemmy-ansible repo, or somewhere else?

Feel free to close this PR, then.

Nutomic:
In the installation page of lemmy-docs repo probably

PR #90, opened 2023-06-21
Add a 2nd nginx layer

dessalines:
I tested this to make sure, and its working on a fresh install. 

This adds a 2nd nginx layer, `nginx_internal.conf`, identical to the one in our lemmy docker folder.

It then adds a system nginx, which is a very simple reverse proxy. We can link to both of these directly in future docker install docs as well.

This will also help people with docker installs, using other simple reverse proxies.

Once this is merged, I'll update our install docs.


templates/docker-compose.yml
```diff
@@ -1,54 +1,100 @@
-version: '2'
+version: "3.3"
```
dessalines, line 1: I changed this around to more closely match our dockerfile in lemmy.

templates/nginx.conf
dessalines, line 1: This file is now extremely simple, just a nginx reverse proxy.

templates/nginx_internal.conf
```diff
@@ -0,0 +1,72 @@
+worker_processes 1;
```
sunaurus, line 1: Is there any reason to not just use `auto` here?

sunaurus:
Just want to highlight some trade-offs of this apporach:

1. I have a feeling that this could end up increasing the setup complexity for all admins, and create additional questions which need to be answered, like: "Which nginx config do I need to change?" or "I increased the `client_max_body_size` in my nginx.conf, why are uploads still limited to 20 mb?", etc
2. Having an extra nginx process + an extra docker container will require some small additional resources + small additional latency for every single HTTP request

templates/docker-compose.yml
Nutomic: FYI its 3.7 now, but apparently the version doesnt matter anyway.

https://github.com/LemmyNet/lemmy/pull/3248

templates/docker-compose.yml
```diff
@@ -1,54 +1,100 @@
...
       - LEMMY_UI_HTTPS=true
     volumes:
       - ./volumes/lemmy-ui/extra_themes:/app/extra_themes
-    depends_on: 
+    depends_on:
       - lemmy
+    restart: always
+
+  pictrs:
+    image: asonix/pictrs:0.4.0-beta.19
```
Nutomic, line 58: Latest version is [v0.4.0-rc.7](https://git.asonix.dog/asonix/pict-rs/releases/tag/v0.4.0-rc.7)

templates/docker-compose.yml
```diff
@@ -1,54 +1,100 @@
...
+      - PICTRS__MEDIA__GIF__MAX_AREA=65536
+      - PICTRS__MEDIA__GIF__MAX_FRAME_COUNT=400
+    user: 991:991
+    volumes:
+      - ./volumes/pictrs:/mnt:Z
+    restart: always
+    deploy:
+      resources:
+        limits:
+          memory: 690m
```
Nutomic, line 82: I think we should also set log limits for all containers like here:

https://github.com/LemmyNet/lemmy/blob/main/docker/docker-compose.yml#L3

templates/nginx_internal.conf
dessalines: Gotcha, I'll do that.

dessalines:
@sunaurus 

> Having an extra nginx process + an extra docker container will require some small additional resources + small additional latency for every single HTTP request

I def agree, I was against a 2-layered setup for a while. But people really couldn't figure out how to translate the full nginx setup into each of their preferred system webservers. Also I think someone showed us that the latency of this multilayered setups was minimal.

> I have a feeling that this could end up increasing the setup complexity for all admins, and create additional questions which need to be answered, like: "Which nginx config do I need to change?" or "I increased the client_max_body_size in my nginx.conf, why are uploads still limited to 20 mb?", etc

We can address those as they come, but I think that specific limit is a good one to have in the internal, as opposed to the external setup.

Nutomic:
The setup would be much cleaner if it wasnt necessary to copy the internal nginx config to the server. Another way would be to prebuild a docker image with nginx and that config, but that would be a bit more complicated to setup.

Im not worried about the resource usage, nginx is extremely lightweight so this wouldnt be noticable.

dessalines:
Its just one more `wget`. After this is merged, I'll make a PR, and test the docker install instructions too.

Nutomic (approved):

kirpy0x:
Thanks for this! I was trying a few days to get it working with my Caddy reverse proxy. Slowly got it working more and more and eventually got all the Test curls returning JSON but still couldn't federate. After seeing this update I pulled, updated the caddy and voila.

PR #92, opened 2023-06-22
Add Ubuntu dependency to README

### LINKED ISSUE #81: Update documentation to include Debian requirement
This playbook assumes you're running on a Debian based linux distro. The readme only calls out "A VPS" as a requirement without specifying that it won't work in distros that don't use aptitude.

Zetaphor:
This resolves #81 to include the Ubuntu dependency.


dessalines:
Debian based, not ubuntu.

dessalines (approved):

PR #93, opened 2023-06-22
Forward host headers from external to internal nginx

Rulasmur:
This fixes issue raised here: https://github.com/LemmyNet/lemmy/issues/3273

IP and host were not being forwarded, causing http signature verification to fail with message `Incoming activity has invalid signature`


dessalines (approved):
Thanks, sorry I forgot about these ones.

PR #99, opened 2023-06-23
Update docker-compose.yml

XIRGO:
Updated docker-compose file as the postfix service needed the ability to access the lemmyinternal network and the lemmyexternal network. This allows the lemmy container to be able to reach the `postfix` dns name. The external network allows for it to be able to ping and resolve external hostnames.

### DIFF
--- a/templates/docker-compose.yml
+++ b/templates/docker-compose.yml
@@ -110,6 +110,9 @@ services:
 
   postfix:
     image: mwader/postfix-relay
+    networks:
+      - lemmyinternal
+      - lemmyexternalproxy
     environment:
       - POSTFIX_myhostname={{ domain }}
     restart: "always"


ross-spencer:
Thanks for this @XIRGO works for me.

nicolevanderhoeven:
Hmmm. I update my docker-compose and redeployed, but it's still not working for me. I do get a "Sent an email to reset your password" message now when I click "forgot your password", but I still don't get an email. Anything else I'm missing? (I just set this up for the first time so apologies if I've missed a crucial step.)

ross-spencer:
@nicolevanderhoeven you might need to test email locally on the server fist, e.g. `echo "test mail." | mail -s "test email subject" -a "From: <you@yourdomain.com>" test_receiver@big_mail_app.com`

(You can also check mail locally by typing `mail` and you may get some good debug help if the mail is returned)

But if that email doesn't arrive either in your inbox or spam you might need to do the following: 

* configure postfix: https://phoenixnap.com/kb/postfix-smtp
* create an SPF record for your domain so that emails aren't blocked.

dessalines:
I've removed those pointless networks now.

PR #102, opened 2023-06-24
Clean up networks in docker-compose.yml

christopher-conley:
This commit removes the "internal" boolean from the "lemmyinternal" network, explicitly defines hostnames for each container (matching existing hostnames if found elsewhere like in env variables), ensures the nginx container connects to the "lemmyexternalproxy" network first, and places the containers in the appropriate network(s).

I'm really not sure what the "internal" boolean on the "lemmyinternal" network brings, as the containers that need WAN access wouldn't be able to get it unless they're _also_ connected to the "lemmyexternalproxy" network and could egress through that route, which defeats the purpose of having them connected to a restricted network. If the idea is to restrict access/traffic to only the containers within that network, but still allow WAN egress from them, it might make more sense to have the internal nginx container also act as a forward proxy for those containers in addition to the reverse proxy role it's already serving.

Not every container had an explicit hostname, so sometimes traffic meant for one container would actually go to a different one. The postfix container also didn't have an explicit network assignment, so it would fall off onto its own subnet in the default network.

### DIFF
--- a/templates/docker-compose.yml
+++ b/templates/docker-compose.yml
@@ -7,18 +7,22 @@ x-logging: &default-logging
     max-file: "4"
 
 networks:
-  # communication to web and clients
-  lemmyexternalproxy:
+  default:
   # communication between lemmy services
   lemmyinternal:
     driver: bridge
-    internal: true
+  # communication to web and clients
+  lemmyexternalproxy:
+    driver: bridge
 services:
   proxy:
     image: nginx:1-alpine
+    hostname: proxy
     networks:
-      - lemmyinternal
-      - lemmyexternalproxy
+      lemmyinternal:
+        priority: 10
+      lemmyexternalproxy:
+        priority: 1000
     ports:
       # actual and only port facing any connection from outside
       # Note, change the left number if port 1236 is already in use on your system
@@ -37,7 +41,6 @@ services:
     hostname: lemmy
     networks:
       - lemmyinternal
-      - lemmyexternalproxy
     restart: always
     logging: *default-logging
     environment:
@@ -50,9 +53,9 @@ services:
 
   lemmy-ui:
     image: {{ lemmy_docker_ui_image }}
+    hostname: lemmy-ui
     networks:
       - lemmyinternal
-      - lemmyexternalproxy
     environment:
       - LEMMY_UI_LEMMY_INTERNAL_HOST=lemmy:8536
       - LEMMY_UI_LEMMY_EXTERNAL_HOST={{ domain }}
@@ -72,7 +75,6 @@ services:
     # entrypoint: /sbin/tini -- /usr/local/bin/pict-rs -p /mnt -m 4 --image-format webp
     networks:
       - lemmyinternal
-      - lemmyexternalproxy
     environment:
       - PICTRS_OPENTELEMETRY_URL=http://otel:4137
       - PICTRS__API_KEY=API_KEY
@@ -109,6 +111,8 @@ services:
     logging: *default-logging
 
   postfix:
+    networks:
+      - lemmyinternal
     image: mwader/postfix-relay
     environment:
       - POSTFIX_myhostname={{ domain }}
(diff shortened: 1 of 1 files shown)


templates/docker-compose.yml
```diff
@@ -72,7 +75,6 @@ services:
     # entrypoint: /sbin/tini -- /usr/local/bin/pict-rs -p /mnt -m 4 --image-format webp
     networks:
       - lemmyinternal
-      - lemmyexternalproxy
```
ticoombs, line 75: pictrs needs to be able to connect outbound. It caches external images. If removing access to this network removes outbound connectivity, then this would break pictrs fetching images.

templates/docker-compose.yml
```diff
@@ -37,7 +41,6 @@ services:
     hostname: lemmy
     networks:
       - lemmyinternal
-      - lemmyexternalproxy
```
ticoombs, line 40: Is it Lemmy or Lemmy UI that sends notifications to other federated instances? Wouldn't we need the "external" network to send federated items to them? (posts/comments/etc)

templates/docker-compose.yml
```diff
@@ -7,18 +7,22 @@ x-logging: &default-logging
...
   proxy:
     image: nginx:1-alpine
+    hostname: proxy
     networks:
-      - lemmyinternal
-      - lemmyexternalproxy
+      lemmyinternal:
+        priority: 10
+      lemmyexternalproxy:
+        priority: 1000
```
ticoombs, line 25: An explanation on the priority setting. Is there a reason why you chose 1000 instead of 100, or 20?

templates/docker-compose.yml
christopher-conley: Gotcha, I had originally started making this commit during the 0.18 rollout when some things were broken immediately after. I'll amend the PR to reflect the desired state.

templates/docker-compose.yml
christopher-conley: From my testing just now, it's Lemmy. I'll update the PR and add the "internal" flag back to the "lemmyinternal" network and put the "lemmy" container in both.

templates/docker-compose.yml
christopher-conley: It was a typo, meant to be "100." From a functional/technical standpoint, there's no detrimental effect having it that high or having that sort of delta between them. I'll adjust it down in the PR edit.

christopher-conley:
I've made a new commit to the PR addressing the comments. I also tested it to a new VM at a VPC, and it deployed and federated with other instances successfully.

Nutomic:
Honestly I would just get rid of all the network sections. They dont seem to add any value, as every service needs to connect to the external network in one case or another. Better to keep it simple and use the default docker network.

christopher-conley:
> Honestly I would just get rid of all the network sections. They dont seem to add any value, as every service needs to connect to the external network in one case or another. Better to keep it simple and use the default docker network.

I think we can close this PR, it looks like that was already done with this commit: [6ebe1a1](https://github.com/LemmyNet/lemmy-ansible/commit/6ebe1a1fe5d1b92f3091662a5e17464f494a6aeb). One downside to this approach is if someone has multiple docker compose files in the same directory, the containers launched from those different compose files will all be connected to the same network. If networks aren't defined and services assigned to them explicitly in the compose files, docker will create a default network for it based on the parent directory name. Has the potential to cause problems in the future.

Also, if the idea was to keep the lemmy/lemmy-ui/postgres containers segregated on their own network and have the nginx container be the only point of network ingress/egress for them, I'm not sure if the internal nginx container serves a purpose after removal of the docker networks. With the networks gone, it looks like that container could be removed entirely and just move the "nginx_internal.conf" config into one of the host's nginx include directories with minor alterations (replace references to lemmy/lemmy-ui with `localhost` and port, essentially).

Nutomic:
I would say t he internal nginx is mainly to make configuration easier.

Issue #106, opened 2023-06-25
[Bug]: Unable to fetch status on some instances (especially, Mastodon)


perillamint:
### Summary

Currently, Lemmy's `nginx_internal.conf` only redirects requests to `lemmy` if

* `$http_accept` has exactly the same value `application/activity+json` or `application/ld+json; profile="https://www.w3.org/ns/activitystreams"`
* `$request_method` is `POST`

However, some AP implementations (Discovered on Mastodon) use multiple mime-types in `Accept` header. For example,

```
Accept: application/activity+json, application/ld+json; profile="https://www.w3.org/ns/activitystreams", text/html;q=0.1
```

and this causes federation problem with Mastodon and other instances.

### Steps to Reproduce

1. Deploy Lemmy instance
2. Post random status on community
3. Run `curl -v -H 'Accept: application/activity+json, application/ld+json; profile="https://www.w3.org/ns/activitystreams", text/html;q=0.1' <your lemmy status URL here>`

Expected behavior: Server returns proper activity+json payload from `lemmy`
Actual behavior: Server returns HTML response from `lemmy-ui`

### Version

BE 0.18.0, commit 63d3759c481ff2d7594d391ae86e881e2aeca56d

### Lemmy Instance URL

N/A

ticoombs:
I cannot replicate this issue on 3 servers that I tested. 
- 0.17.4
- 0.18.0

They all returned json data. I even tried swapping the accept headers to be `application/ld+json, application/activity+json;` and it still worked. 

So not sure what error you are getting. Can you please double check that you are still getting the error on the initial server you were testing?

perillamint:
Can you test it with `https://<lemmy.url>/post/<post number>`? For me, it responds

```
$ curl -H 'Accept: application/activity+json, application/ld+json; profile="https://www.w3.org/ns/activitystreams", text/html;q=0.1' https://<my lemmy tunnel url>/post/6    

    <!DOCTYPE html>
    <html lang="en">
    <head>
    <script>window.isoData = {"
```

and if I modify Accept to `application/activity+json`

```
$ curl -H 'Accept: application/activity+json' https://<my lemmy tunnel url>/post/6     
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    "https://w3id.org/security/v1",
    {
      "lemmy": "https://join-lemmy.org/ns#",
      "litepub": "http://litepub.social/ns#",
      "pt": "https://joinpeertube.org/ns#",

[...]

PR #107, opened 2023-06-25
Use nginx regexp to accept multiple Accept header content

perillamint:
This patch utilizes regular expression to attempt properly route in the case of multiple accept header content.

Fixes issue #106


ticoombs:
Confirming this works.

ticoombs (approved):

jippi:
I got an alternative solution using the native `map` in nginx here https://github.com/LemmyNet/lemmy-ansible/issues/106#issuecomment-1606222766

Issue #82, opened 2023-06-12
Problems on Ubuntu 23.04: Unneccessary(?) python modules breaking the installation, deprecated apt-keyring


darklajid:
Tried getting this up and running on a brand new VPS running 23.04

It failed during the "Install Docker Module and docker-compose for Python" step - `pip3 install -U docker docker-compose` gave a scary warning about mixing system-managed packages with pip and refused to continue. I worked around this by removing this whole part in the playbook: It seems to be merely required for the playbook itself, to run a single docker compose command.

* It fails with my system out-of-the-box
* The docker_compose Ansible plugin seems outdated (requires the legacy `docker-compose` < 2.x instead of the > 2.x `docker compose`)
* It seems not worth pulling in dependencies for a simple, single command

I changed the lemmy.yml with the following diff to make things work:

```
diff --git a/lemmy.yml b/lemmy.yml
index 2c6a7aa..41fa786 100644
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -59,13 +59,6 @@
         state: latest
         update_cache: true
 
-    - name: Install Docker Module and docker-compose for Python
-      pip:
-        name:
-          - docker
-          - docker-compose
-        state: latest
-
     - name: copy docker config
       copy: src='../files/docker-daemon.json' dest='/etc/docker/daemon.json' mode='0644'
 
@@ -160,11 +153,9 @@
     # - debug: var=shell_output
 
     - name: start docker-compose
-      docker_compose:
-        project_src: "{{lemmy_base_dir}}/{{domain}}"
-        state: present
-        pull: yes
-        remove_orphans: yes
+      shell:
+        cmd: docker compose up -d
+        chdir: "{{lemmy_base_dir}}/{{domain}}"
 
     - name: reload nginx with new config
       shell: nginx -s reload

```

After this I now get warnings running apt update:

```
W: https://download.docker.com/linux/ubuntu/dists/focal/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
```

Arguably that's yet another [outdated plugin](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html) that's doing something it shouldn't.

From the link above / in the Ansible docs:

[...]

Issue #61, opened 2023-03-09
Error on the Task - Add Docker GPG apt Key


admiralarjun:
TASK [Add Docker GPG apt Key] ***************************************************************

`fatal: [admin@<IP>]: FAILED! => {"changed": false, "cmd": "/usr/bin/apt-key add -", "msg": "Unable to add a key from binary data", "rc": 2, "stderr": "Warning: apt-key output should not be parsed (stdout is not a terminal)\ngpg: failed to start agent '/usr/bin/gpg-agent': No such file or directory\ngpg: can't connect to the agent: No such file or directory\n", "stderr_lines": ["Warning: apt-key output should not be parsed (stdout is not a terminal)", "gpg: failed to start agent '/usr/bin/gpg-agent': No such file or directory", "gpg: can't connect to the agent: No such file or directory"], "stdout": "", "stdout_lines": []}`

Nutomic:
Sounds like you need to install gpg or gpg-agent package.

Nutomic:
You can also install docker-compose from official Ubuntu repos. To be honest I don't think there is any reason to use official Docker repository, especially if it causes problems like this.

admiralarjun:
> You can also install docker-compose from official Ubuntu repos. To be honest I don't think there is any reason to use official Docker repository, especially if it causes problems like this.

Can i get the modified version of the playbook with this changes?

Nutomic:
I made the change in #62. Let me know if it works with that.

admiralarjun:
Thank You, anyway I managed to perform the installation by manually installing the docker and commenting the docker steps in the playbook.

dessalines:
> stderr_lines": ["Shared connection to <ip>closed."], "stdout": "\u001b[33m\r0% [Working]\u001b[0m\r          
\rGet:1 https://download.docker.com/linux/ubuntu xenial InRelease [66.2 kB]\r\n\u001b[33m\r0%
[Connecting to archive.ubuntu.com]

I googled that error, and it looks like it has to do with either an old version of ansible on your system, or an old version of python on the target.

https://www.tecmint.com/fix-shared-connection-to-x-x-xx-closed-ansible-error/

admiralarjun:
And, does Lemmy maintain any error logs? My instance's emails are not working. What could possibly go wrong? and I don't remember giving any SMPT details during the installation nor the setup

admiralarjun:
I did my installation with ansible, where I make the changes now?

[...]

Issue #54, opened 2022-09-20
Unable to apdate


DokStook:
When I try to update using the steps mentioned in the README I get the error 

```

fatal: [redacted@redacted]: FAILED! => {"changed": false, "cmd": "apt-get install --no-install-recommends python-apt -y -q", "msg": "E: Package 'python-apt' has no installation candidate", "rc": 100, "stderr": "E: Package 'python-apt' has no installation candidate\n", "stderr_lines": ["E: Package 'python-apt' has no installation candidate"], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nPackage python-apt is not available, but is referred to by another package.\nThis may mean that the package is missing, has been obsoleted, or\nis only available from another source\nHowever the following packages replace it:\n  python3-apt-dbg python3-apt python-apt-doc python-apt-common\n\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "Package python-apt is not available, but is referred to by another package.", "This may mean that the package is missing, has been obsoleted, or", "is only available from another source", "However the following packages replace it:", "  python3-apt-dbg python3-apt python-apt-doc python-apt-common", ""]}

```

Looking at it I can see that there are other packages that are suggested instead of 'python-apt'. Should this be changed somewhere in the config?

dessalines:
I don't know what distro you're using, but it looks like they removed python2.

Check the last line in these install instructions: https://github.com/LemmyNet/lemmy-ansible#install

BrienJoseph:
https://docs.ansible.com/ansible/latest/reference_appendices/python_3_support.html

I would try to declare the python interpreter to your version of python. Had some trouble w/ this on porting over things from rhel7 to rhel8 on a different project via ansible playbooks.

eslerm:
I reproduced this issue on Ubuntu 23.04 (Lunar). Ubuntu 22.04 LTS (Jammy) is not affected.

dessalines:
Stale issue. Feel free to re-open if its still a problem.

PR #116, opened 2023-07-01
configure jammy++ docker binaries

eslerm:
lemmy-ansible uses deprecated `apt-key` and Docker's apt repo is only configured for Ubuntu 20.04 (Focal) binaries.

To reproduce run lemmy-ansible on an Ubuntu 22.04 (Jammy) vm. Login and check apt configuration:
```shell
eslerm@sec-jammy-amd64:~$ sudo cat /etc/apt/sources.list.d/download_docker_com_linux_ubuntu.list
deb https://download.docker.com/linux/ubuntu focal stable
```
```shell
eslerm@sec-jammy-amd64:~$ sudo apt update
Hit:1 http://archive.ubuntu.com/ubuntu jammy InRelease
Hit:2 http://security.ubuntu.com/ubuntu jammy-security InRelease
Hit:3 https://download.docker.com/linux/ubuntu focal InRelease
Hit:4 http://archive.ubuntu.com/ubuntu jammy-updates InRelease
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
35 packages can be upgraded. Run 'apt list --upgradable' to see them.
W: https://download.docker.com/linux/ubuntu/dists/focal/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
N: Skipping acquire of configured file 'stable/binary-i386/Packages' as repository 'https://download.docker.com/linux/ubuntu focal InRelease' doesn't support architecture 'i386'
```

This commit makes apt configuration conditional on release number. The pre-Jammy configurations were left as is. Jammy++ configs adapt to ansible_distribution_release (i.e., jammy, kinetic, lunar, and mantic get the correct binaries) and use the `signed-by` apt config, which replaces `apt-key` and is [suggested by Docker](https://docs.docker.com/engine/install/ubuntu/). I did not dearmor the key as Docker suggested for Ansible convinence. Please verify the checksum. The README.me says that Debian-based vms are supported, but that's was not the case for Docker binaries before or with this commit.

I tested the patch on Focal, Jammy, and Lunar. Focal and Lunar failed after apt config unrelated preexisting issues (https://github.com/LemmyNet/lemmy-ansible/issues/54 in the case of Lunar). This fixes https://github.com/LemmyNet/lemmy-ansible/issues/61 and partially resolves https://github.com/LemmyNet/lemmy-ansible/issues/82

### DIFF
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -45,7 +45,7 @@
           - "python3-setuptools"
 
     - name: Configure Docker apt repo before Jammy
-      #when: ansible_distribution == 'Ubuntu' and ansible_distribution_version < '22.04'
+      when: ansible_distribution == 'Ubuntu' and ansible_distribution_version < '22.04'
       block:
         - name: Add Docker GPG apt Key
           apt_key:
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -38,20 +38,36 @@
           - "apt-transport-https"
           - "ca-certificates"
           - "curl"
+          - "gnupg"
           - "software-properties-common"
           - "python3-pip"
           - "virtualenv"
           - "python3-setuptools"
 
-    - name: Add Docker GPG apt Key
-      apt_key:
-        url: https://download.docker.com/linux/ubuntu/gpg
-        state: present
+    - name: Configure Docker apt repo before Jammy
+      #when: ansible_distribution == 'Ubuntu' and ansible_distribution_version < '22.04'
+      block:
+        - name: Add Docker GPG apt Key
+          apt_key:
+            url: https://download.docker.com/linux/ubuntu/gpg
+            state: present
+        - name: Add Docker Repository
+          apt_repository:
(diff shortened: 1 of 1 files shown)


eslerm:
please squash :)

Nutomic (approved):

Nutomic:
Thanks! This would be so much easier if we just used Docker from Ubuntu repos...

eslerm:
> This would be so much easier if we just used Docker from Ubuntu repos...

After a new Ubuntu Release starts, most packages in the previous release are version locked and only receive security updates except for extraordinary circumstances. That's why Docker suggests using their repos. Using the docker snap could fit the same need though.

With this merge, the Docker repos are now only added ` when: ansible_distribution == 'Ubuntu'`. Please be aware that Debian and other distros need explicit support now (the merge removed implicit support using incorrect configs).

PR #118, opened 2023-07-03
Adding some more custom postgres params.

dessalines:


Nutomic (approved):

Nutomic:
Ive deployed these to lemmy.ml

PR #124, opened 2023-07-08
Add Traefik deployment and prepare for lemmy-docs update

Kangie:


TheSilkky:
Based on [this](https://github.com/LemmyNet/lemmy-ansible/blob/main/templates/nginx_internal.conf) wouldn't this rule would make more sense?
```
traefik.http.routers.lemmy.rule=Host(`${HOST}`) && PathPrefix(`/api`, `/pictrs`, `/feeds`, `/nodeinfo`, `/.well-known`) || (Method(`GET`, `HEAD`) && HeaderRegexp(`Accept`, `^application/(activity|ld)\+json$`)) || !Method(`GET`, `HEAD`)
```

Nutomic:
I would really prefer to keep this repo simple, with a basic docker-compose + nginx deployment. Adding other deployment options makes it more complicated to maintain, as changes need to be applied to all variants. Its especially problematic that neither I nor dessalines have ever used traefik before, so we are unable to review changes properly, meaning things will likely keep breaking. And there is no guarantee that you or other contributors will still be around in a year or two.

So traefik installation instructions would better go into your own git repo, blog post or lemmy post, and then you can add a link to that in the lemmy docs.

dessalines:
Agree, other non-official installs should go in the lemmy-docs, or on your own repo.

Kangie:
@TheSilkky I updated the other PR; let's discuss your proposed changes over on the GitLab repo that I've linked to containing the traefik configuration :)

PR #126, opened 2023-07-08
Link to github releases page instead of RELEASES.md

hpincket:
This change updates a link in README.md so that it points to the github releases page instead of RELEASES.md. It doesn't look as if RELEASES.md is kept up to date. RELEASES.md was last updated 5 months ago.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ To run this ansible playbook, you need to:
 ## Upgrading
 
 - Run `git pull`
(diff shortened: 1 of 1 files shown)


Nutomic:
I would rather use one of these as the main places to get announcements:
- https://lemmy.ml/c/announcements
- https://join-lemmy.org/news

dessalines:
No, because if / when we migrate off github, that releases endpoint might not exist on the new platform.

PR #129, opened 2023-07-10
Add python script to generate configs

adambowles:
Move docker_compose.yml to examples
Install python3 dependencies via pipenv
Run python install script

Script will:
Optionally clean up previous runs
Copy `docker_compose.yml` to `templates/`
Ask for all the required information
Ask whether to use object storage, and if yes ask for object storage details

Example run:

image


Nutomic:
So the Python script is only used to provide an interactive way to set config values? I dont think we want to add this much extra complexity. Its better if people edit config files directly, then they know where to change things if something is wrong. This script makes things very intransparent.

adambowles:
Okay 🙂

Nutomic:
The change for pictrs env vars is still needed though. Could you make a PR for that?

adambowles:
I would but that's the only part I didn't get working.. ansible is not something I have any experience in and I got ignored in their support discord so..

PR #136, opened 2023-07-22
Advanced postfix container config

padthaitofuhot:
This PR adds advanced configuration capabilities for the Postfix relay container by way of an additional host vars file and changes to docker-compose.yml.

The specific use-case is detailed in _[Bug]: Cannot send email using external SMTP server [#3348](https://github.com/LemmyNet/lemmy/issues/3348)._

README.md now contains documentation for using additional host vars.  An additional host vars file has been added to `examples/`.

This PR has been tested both using and not using the additional host vars to configure Postfix.  The default Postfix configuration (without using additional host vars) retains behavior identical to behavior on the main branch.


padthaitofuhot:
Sorry, I completely wrecked this PR lol

PR #144, opened 2023-08-07
Almalinux support

codyro:


codyro:
Whoops, I didn't mean to make this PR yet. I need to do some more testing first!

Clearly bedtime though :).

Issue #123, opened 2023-07-07
Changes to docker apt repository broke non-ubuntu instances


snowe2010:
We're completely unable to install using lemmy-ansible due to the changes in commit d2403a1976c7c21241cbd39ebad66b70df369ca6 which force usage of Ubuntu.

eslerm:
Before https://github.com/LemmyNet/lemmy-ansible/commit/d2403a1976c7c21241cbd39ebad66b70df369ca6 all installs used the 20.04 binary of docker.

Which distro do you need support for?

ismailkarsli:
> Before [d2403a1](https://github.com/LemmyNet/lemmy-ansible/commit/d2403a1976c7c21241cbd39ebad66b70df369ca6) all installs used the 20.04 binary of docker.
> 
> Which distro do you need support for?

I'm using Debian 11 and getting the error above:
```TASK [Download Docker GPG Key] *******************************************************************************************
fatal: [root@128.140.48.219]: FAILED! => {"changed": false, "checksum_dest": null, "checksum_src": "f5b5bd1487cefc0c53c947e11ca202e86b33dbad", "dest": "/etc/apt/keyrings/docker.asc", "elapsed": 0, "msg": "Destination /etc/apt/keyrings does not exist", "src": "/root/.ansible/tmp/ansible-moduletmp-1691581922.179723-24vaqrm8/tmp8hlvs__1", "url": "https://download.docker.com/linux/debian/gpg"}```

snowe2010:
sorry, I somehow missed the notification. I'm also using Debian 11. I had to completely revert the commit that changed that. It wasn't due to the docker image being ubuntu, it was like @ismailkarsli said, the when stopped this step from ever being run.

snowe2010:
why have the check for Ubuntu (or as you're suggesting, Debian) at all? Everything would continue to work fine without that check. Until logic is added for other non-apt systems it's just breaking all the other apt systems from working with lemmy-ansible.

codyro:
The `lemmy-ansible` project should have a few target distributions that it supports to make maintaining and supporting easier to avoid running into issues like this. 

I suggest starting by offering the following targets and modifying the layout to be a bit more modular (EX; https://github.com/LemmyNet/lemmy-ansible/pull/145#issuecomment-1671639359, or similar):

* Ubuntu 22.04 LTS
* AlmaLinux/RHEL/CentOS/RockyLinux 9
* Debian 10/11

I can assist in this.

[...]

PR #148, opened 2023-08-09
Get dpkg-formatted arch and use it for docker apt repo

### LINKED ISSUE #130: Ansible setup has hard dependency on amd64
The following hard-coded instance of `amd64` means that installation of Docker from apt repository will fail on devices like RPI 4.

https://github.com/LemmyNet/lemmy-ansible/blob/41f5f55cf70d4d7790466566d1dc0eafbc1be768/lemmy.yml#L69C30-L69C35

This will result in the following output when trying to `sudo apt install docker-ce`:

```
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package docker-ce is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  docker-ce-cli:amd64

E: Package 'docker-ce' has no installation candidate
```

codyro:
Should pass CI checks, tested against Jammy amd64/arm64. Ansible has no built-in facts that return the format that `apt`/`dpkg` is looking for (EX; `ansible_architecture`).

- Fixes #130
- Tested against amd64/arm64

### DIFF
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -57,6 +57,8 @@
             state: present
 
     - name: Get architecture using dpkg
+      when: (ansible_distribution == 'Debian') or
+        (ansible_distribution == 'Ubuntu' and ansible_distribution_version >= '22.04')
       command: dpkg --print-architecture
       register: dpkg_output
       changed_when: false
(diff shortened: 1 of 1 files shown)


codyro:
Added small tweak that addresses/resolves #123

snowe2010:
Tested and worked for programming.dev

Nutomic:
@iheart2code Does this work for you?

@snowe2010 Which architecture does programming.dev use?

Harrisoned:
Can confirm it working on a Debian GNU/Linux 11 (bullseye).

omalaspinas:
Worked for me too.

dessalines (approved):
Cool, thx!

snowe2010:
@Nutomic sorry, somehow missed your tag. programming.dev is on `x86_64`, though I'm not sure that's exactly what you're asking?

PR #150, opened 2023-08-11
Fix frontend rewrite on internal nginx instance

### LINKED ISSUE #149: Strange redirect rule on slash at the end of URL
Go to `https://lemmy.ml/c/announcements` and everything works as you expect.

However, if you go to `https://lemmy.ml/c/announcements/` (notice the added slash at the end), you are redirected to `https://lemmy.ml:8536/c/announcements`, which fails to load.

I'm guessing this has something to do with this rule in the internal nginx configuration:
```
        # frontend general requests
        location / {
            proxy_pass $proxpass;
            rewrite ^(.+)/+$ $1 permanent;
        }
```

But I'm not sure. Also not sure how to fix it. Ideally it should redirect to `https://lemmy.ml/c/announcements` without the specified port.

codyro:
- Fixes #149

### DIFF
--- a/templates/nginx_internal.conf
+++ b/templates/nginx_internal.conf
@@ -73,8 +73,11 @@ http {
 
         # frontend general requests
         location / {
-            proxy_pass $proxpass;
-            rewrite ^(.+)/+$ $1 permanent;
+            # If there's a trailing slash, rewrite the URL without redirecting
+            rewrite ^(.+)/+$ $1 break;
+
+            # Then, proxy to the appropriate upstream based on $proxpass
+            proxy_pass $proxpass;           
         }
 
         # security.txt


Nutomic:
I tried this on lemmy.ml but it didnt work. The problem is that it redirects to the internal address, so `https://lemmy.ml/c/announcements/` becomes `https://lemmy:8536/c/announcements`. Moving the rewrite rule from docker internal nginx to system nginx fixes the redirect.

codyro:
This was tested in the parent issue #149 on my test instance. I just verified it works once more. 

Can you please let me know what you're seeing when you change this rewrite rule? I can't replicate the internal redirect using my changes.

codyro:
Are you looking for it to strip trailing slashes? Is there a functional reason, or is it mostly aesthetic? 

I can just adjust the PR to go along @Nutomic's suggestions. I'll do some testing in a bit, and assuming it looks okay, I'll commit to this branch/PR :)

dessalines (commented):
Seems like redirecting it would be a better solution.

PR #151, opened 2023-08-11
Fix duplicate headers being sent by nginx

### LINKED ISSUE #143: Duplicate headers between external and internal nginx configs
Hi, unsure if this is an issue, but I noticed that my Lemmy instance was getting flagged on the Mozilla Observatory for unrecognized X-Content-Type-Options, X-Frame-Options and X-XSS-Protection headers which is weird since it definitely is in the config on this repo.

Looking at what it sees when it scans my site I realized that these headers appear in both configs and this is what is tripping up Mozilla. I don't know if this matters in practice.

Example headers after upgrade to 0.18.3:
x-content-type-options: nosniff, nosniff
x-frame-options: SAMEORIGIN, DENY
x-xss-protection: 1; mode=block, 1; mode=block

I can submit a PR modifying the config, but would it be better to remove these headers from the internal or external config? It seems to only matter for the X-Frame-Options header.

codyro:
The only noteworthy change is I opted to send `X-Frame-Options: DENY` instead of `X-Frame-Options: SAMEORIGIN` as it's more restrictive, and I don't believe Lemmy needs anything more lax. It is also in line with the documentation instructions from [install from scratch](https://join-lemmy.org/docs/administration/from_scratch.html#configure-reverse-proxy-and-tls), which uses the [nginx.conf](https://raw.githubusercontent.com/LemmyNet/lemmy-ansible/main/templates/nginx.conf) from this repository, which uses `X-Frame-Options: DENY`.

**Before**
```
vary: Accept-Encoding
vary: Accept-Encoding
x-content-type-options: nosniff
x-content-type-options: nosniff
x-xss-protection: 1; mode=block
x-xss-protection: 1; mode=block
x-frame-options: SAMEORIGIN
x-frame-options: DENY
```

**After**
```
vary: Accept-Encoding
x-content-type-options: nosniff
x-xss-protection: 1; mode=block
x-frame-options: DENY
```

- Internal/external nginx were sending duplicate headers, removed from internal
- Opted X-Frame-Options DENY as default
- Fixes #143

### DIFF
--- a/templates/nginx_internal.conf
+++ b/templates/nginx_internal.conf
@@ -57,14 +57,9 @@ http {
 
         gzip on;
         gzip_types text/css application/javascript image/svg+xml;
-        gzip_vary on;
 
         # Upload limit, relevant for pictrs
         client_max_body_size 20M;
-
-        add_header X-Frame-Options SAMEORIGIN;
-        add_header X-Content-Type-Options nosniff;
-        add_header X-XSS-Protection "1; mode=block";
 
         # Send actual client IP upstream
         proxy_set_header X-Real-IP $remote_addr;
--- a/templates/nginx_internal.conf
+++ b/templates/nginx_internal.conf
@@ -55,9 +55,6 @@ http {
         server_name localhost;
         server_tokens off;
 
-        gzip on;
-        gzip_types text/css application/javascript image/svg+xml;
-
         # Upload limit, relevant for pictrs
         client_max_body_size 20M;
 


templates/nginx_internal.conf
```diff
@@ -57,15 +57,10 @@ http {
 
         gzip on;
         gzip_types text/css application/javascript image/svg+xml;
-        gzip_vary on;
```
Nutomic, line 60: You can get rid of the other gzip lines too as the other nginx config already has them.

dessalines: bump

codyro: That shouldn't cause any issues, but I can remove it shortly once I'm done with $dayjob. Thanks for the bump @Nutomic

Nutomic (approved):

PR #153, opened 2023-08-20
Fix pictrs environent variable for setting API_KEY

usr3:
Environment variable for setting API Key has changed in v0.4.0. This fixes the error: Invalid API Key with the old variable.


dessalines (approved):
Thx!

PR #156, opened 2023-09-01
Do a raw test for python3 installation as well

sanatsathaye:


sanatsathaye:
I also plan to make a few PRs to make ansible-lint happier (for most rules but not all because ansible-lint can be stupidly strict at times).

dessalines (approved):
Sweet, thx!

Issue #152, opened 2023-08-13
completely busted on debian 12


marrenia:
this ansible script is completely busted for debian 12 - the first error you get after starting the install is a pip error saying that you cannot install docker-compose the way the script wants - you have to install it as a package using apt. so i fixed that by installing python3-docker using apt but it still doesnt seem to work. maybe look into this?

codyro:
Can you please provide the full output of the error you're receiving? If possible please run ansible with `-vv` for additional verbosity.

You might want to try my changes [here](https://github.com/LemmyNet/lemmy-ansible/pull/148), as it may be related.

stillline:
I had the same error with python and pip. Had to modify the ansible.cfg to install using apt rather than pip

accad:
I am not OP, but I get the same error, although I can get around it, I would prefer if you can come up with a solution for other users

Here is the full error:

TASK [Install Docker Module and docker-compose for Python] **************************************************************************************************
fatal: [user@lemmy.domain.local]: FAILED! => {"changed": false, "cmd": ["/usr/bin/python3", "-m", "pip.__main__", "install", "-U", "docker", "docker-compose"], "msg": "\n:stderr: error: externally-managed-environment\n\n× This environment is externally managed\n╰─> To install Python packages system-wide, try apt install\n    python3-xyz, where xyz is the package you are trying to\n    install.\n    \n    If you wish to install a non-Debian-packaged Python package,\n    create a virtual environment using python3 -m venv path/to/venv.\n    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make\n    sure you have python3-full installed.\n    \n    If you wish to install a non-Debian packaged Python application,\n    it may be easiest to use pipx install xyz, which will manage a\n    virtual environment for you. Make sure you have pipx installed.\n    \n    See /usr/share/doc/python3.11/README.venv for more information.\n\nnote: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.\nhint: See PEP 668 for the detailed specification.\n"}

codyro:
If someone can test the changes in https://github.com/LemmyNet/lemmy-ansible/pull/157, that would help expedite getting this merged in for others :).

PR #157, opened 2023-09-01
Remove reliance on `pip` module 

codyro:
The reliance on the `ansible.builtin.pip` module caused Debian-based systems issues (#152). There are relevant packages in the default repositories of Debian 11, Debian 12, >=Ubuntu 22.04 LTS/they were redundant. As such, I've removed that portion of the playbook to use `apt`. 

After glancing at the repositories for these distributions, I realized that we probably don't need to use the Docker Community Edition and can use what's in the repositories safely. That will be another PR where we clean things up a bit, though :).

I've tested the changes against the targets mentioned above. If someone from #152 can confirm the fix, that'd be fantastic!


dessalines (approved):
Sweet, thx!

Issue #44, opened 2022-05-22
Add LEMMY_UI_CUSTOM_SCRIPT


dessalines:
Add the LEMMY_UI_CUSTOM_SCRIPT to setup.

codyro:
This was renamed to `LEMMY_UI_CUSTOM_HTML_HEADER`, correct? I'll add the environmental variables listed [here](https://github.com/LemmyNet/lemmy-ui/#configuration) to be handled in the same way as `pict-rs` in #160 for the `lemmy-ui` VM.

codyro:
Closing this as #160 will allow any environmental variable to be passed to the lemmy-ui container.

Issue #109, opened 2023-06-26
How to externalize pictrs configuration with ansible?


dessalines:
I'm very much not an ansible expert, so I could really use some help from @Nutomic or others on this one.

Currently pictrs is configured via env-vars, hard-coded into the `docker-compose.yml` . This isn't a good design, because the docker-compose is template file common to all deploys. Some deploys might use an s3 pictrs config, while others, the default file-based one.

Our option are:

- Figure out how to externalize the environment vars provided to pictrs, and include them in `inventory/host_vars/{{instance}}/` somehow.
- Make the `docker-compose.yml` custom for each install. I really don't like this one tho.

mpraeger:
As you are already using ansible's `template` module for docker-compose.yml changing it to be configurable via host_vars seems like something to try before digging into pictrs and how to externalise its configuration is an unknown factor.

`docker-compose.yml` already contains [some pictrs configuration](https://github.com/LemmyNet/lemmy-ansible/blob/main/templates/docker-compose.yml#L57). Making that dynamic and configurable looks pretty straight forward. Which config options were you looking to externalise?

dessalines:
A lot of its `S3` configuration settings, which should be optional. Essentially I need to provide a variable set of env vars to a specific container.

Demigodrick:
You would need to add the following lines to the pictrs environment in docker-compose.yml:
`
      - PICTRS__STORE__TYPE=object_storage
      - PICTRS__STORE__ENDPOINT=<S3 endpoint>
      - PICTRS__STORE__BUCKET_NAME=<bucket name>
      - PICTRS__STORE__REGION=<region>
      - PICTRS__STORE__USE_PATH_STYLE=false 
      - PICTRS__STORE__ACCESS_KEY=<access key>
      - PICTRS__STORE__SECRET_KEY=<secret key>
`
However as dessalines has pointed out, you can't really make them variable because of the setup - two different compose files might be the option, or commenting the above lines out with a note to say only required if using external s3 storage.

adambowles:
Could we use a script to take these configs from one .yml file and add them to the docker compose, such that the end user doesn't have to edit files tracked by git, causing potential issues in upgrades. I know that someone wishing to deploy a Lemmy instance should have a working knowledge of git..

Preferably this script would handle all the deploy steps outlined in the [readme](https://github.com/LemmyNet/lemmy-ansible#install)

adambowles:
Started a PR to do like I suggested:  #129

[...]

PR #160, opened 2023-09-02
Externalize pict-rs configuration, start centralizing configuration variables

### LINKED ISSUE #53: Add PICTRS__API_KEY
Necessary to make admin purge of images work. Needs to be set as pictrs env var in docker-compose.yml, and in lemmy config.

Btw. this also needs to be added to instructions for manual installation.

codyro:
This PR was primarily looking to resolve #53 & #109. However, I ran across some things that required adjusting (which will benefit us when we tidy up everything a bit more!). 

Special thanks to @mpraeger for the [inspiration](https://github.com/LemmyNet/lemmy-ansible/issues/109#issuecomment-1629652836). I've added you as a co-author on the initial commit as I re-used some of your code & idea's. 

### Primary Changes

- Remove the environmental variables defined in the `docker-compose.yml` template for `pict-rs` and move them to a `vars.yml` file which is used to configure & maintain variables in a sane spot (instead of them floating around in the playbook) https://github.com/LemmyNet/lemmy-ansible/commit/a8c859da60d31281faa34d1de57049f728436121
    - Opted to still have `PICTRS__SERVER__API_KEY` default to `postgres_password` to ensure interoperability with prior versions,  but this allows users to have more agency over what it’s set to. https://github.com/LemmyNet/lemmy-ansible/commit/a8c859da60d31281faa34d1de57049f728436121#diff-fadf44a49d340433e809fac78150211bc15b8047e299b14bf891197a2f8a54fdR6
    
- Move `postgres_password` variable to `vars.yml` instead of being defined in `lemmy.yml` and `lemmy-almalinux.yml`. This exposes it to users in case they have an existing database they would like to use or if they’d prefer to set it manually (it also helps clean up the playbooks by centralizing the configuration variables)
    - Rename the generated `inventory/host_vars/<fqdn>/passwords/postgres` to `inventory/host_vars/<fqdn>/passwords/postgres.psk`. The reason for this change is we’re currently not using the `host_vars` path entirely correctly, so it’s attempting to parse the password file as a variable file and will puke due to an invalid syntax. By default, Ansible will read in `inventory/host_vars/<fqdn>/{*.yml,*.yaml,*.ini,exec_file}` files.
- Adjust the example inventory (`examples/hosts`) to use a better format. Instead of using `user@server.com`, we opt to use the `ansible_user=user` to avoid unpredictable variable loading. Using the above examples, if you have `user@server.com` in your inventory, Ansible will look for variables in `inventory/host_vars/user@server.com` directory instead of the expected `inventory/host_vars/server.com` directory.

### Tested against:

- AlmaLinux 9/RHEL 9
- ~Debian 10~ (**UNTESTED**, should work)
- Debian 11
- Debian 12
- Ubuntu 22.04 LTS

Could someone verify at least one or more of the target distributions above? I didn't have time to test `pict-rs` with S3, but I can confirm the environmental variables get populated and injected into the VM properly. We should try this to ensure there is no funny business when using S3 for object storage.


examples/vars.yml
```diff
@@ -0,0 +1,25 @@
...
+  - RUST_BACKTRACE: full
+#  - PICTRS__STORE__TYPE: object_storage 
+#  - PICTRS__STORE__ENDPOINT: '<S3 endpoint>'
+#  - PICTRS__STORE__BUCKET_NAME: '<bucket name>'
+#  - PICTRS__STORE__REGION: '<region>'
+#  - PICTRS__STORE__USE_PATH_STYLE: false 
+#  - PICTRS__STORE__ACCESS_KEY: '<access key>'
+#  - PICTRS__STORE__SECRET_KEY: '<secret key>'
+
+
```
Nutomic, line 23: How about using a .env file instead? Then you can include the vars in docker-compose.yml with a single line `env_file`

templates/docker-compose.yml
```diff
@@ -55,15 +55,13 @@ services:
...
-      - PICTRS__MEDIA__GIF__MAX_WIDTH=256
-      - PICTRS__MEDIA__GIF__MAX_HEIGHT=256
-      - PICTRS__MEDIA__GIF__MAX_AREA=65536
-      - PICTRS__MEDIA__GIF__MAX_FRAME_COUNT=400
+{% if pictrs_env_vars is defined and pictrs_env_vars|length > 0 %}
+{%- for item in pictrs_env_vars %}
+{%- for key, value in item.items() %}
+      - {{ key }}={{ value }}
+{% endfor -%}
+{% endfor %}
```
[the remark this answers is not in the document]
codyro: That is a Jinja2 idiom to control white space. It removes the whitespace from the side of the tag it's on. These were mostly a holdover to get the generated template not to be ugly/conform. However, I don't think it's necessary anymore.

Did you run into any issues with the above code?

examples/vars.yml
codyro: If we go that route, we would still want Ansible to generate the `env_file`(s) accordingly. It allows people to modify and configure the installation in a single centralized place (my goal is to clean this up and get everything in the same spot, then document everything so users have more control over their installations). I think this is a better UX and more "Ansible" oriented.

If this gets merged, I'd do the same thing for `lemmy-ui` immediately and plop all available configuration variables in `vars.yml`, which also resolves #44.  

What are your thoughts on populating the respective `.env` file(s) via Ansible and updating the template to point to the respective `.env` file(s) we generate?

templates/docker-compose.yml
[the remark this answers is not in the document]
codyro: Gave this a look just now. They're now unnecessary with how its indented in the template, so I removed them to clean it up. Nice catch!

examples/vars.yml
```diff
@@ -0,0 +1,25 @@
...
+  - PICTRS__SERVER__API_KEY: "{{ postgres_password }}"
+  - PICTRS__MEDIA__VIDEO_CODEC: vp9
+  - PICTRS__MEDIA__GIF__MAX_WIDTH: 256
+  - PICTRS__MEDIA__GIF__MAX_HEIGHT: 256
+  - PICTRS__MEDIA__GIF__MAX_AREA: 65536
+  - PICTRS__MEDIA__GIF__MAX_FRAME_COUNT: 400
+  - PICTRS_OPENTELEMETRY_URL: http://otel:4137
+  - RUST_LOG: debug
+  - RUST_BACKTRACE: full
+#  - PICTRS__STORE__TYPE: object_storage
```
dessalines, line 15: Thanks a ton for adding these as comments, its going to be really helpful to people who want to use object storage right away.

examples/vars.yml
dessalines: I think the `yml` file is fine, ansible seems to prefer those anyway.

dessalines (approved):
This is great, thanks a ton!

examples/vars.yml
Nutomic: Yes makes sense to move all env vars into a separate file, so that people dont have to edit docker-compose.yml. It also makes it easier to secure passwords by restricting permissions of this file. I would definitely prefer using .env as its much easier (only a single line needed to import).

codyro: The permissions aren't a big concern as the way the playbooks are structured is they're run locally--meaning any sensitive variables stay local to your system. Making a `.env` file `0600` or a `vars.yml` file `0600` is trivial. 

We could use a singular generated .env file for all VMs, but it's a little ugly as it'd populate all the environmental variables for all VMs, which could lead to unexpected results. It also doesn't allow for much flexibility--for example, if I want to set `RUST_LOG` to a different level on two Docker VMs, it wouldn’t be possible with your proposal. 

To do this properly, each VM would have its own environment file. A user must copy the sample environment files to the appropriate directory, edit it, and then run the playbook in contrast to this solution, where a user needs to modify a single file while Ansible does the rest. 

Let me see if I can come up with something that satisfies what you’re looking for without compromising UX.

codyro: What are your thoughts on this approach? This will allow the best mix of UX and flexibility for instance operators. 

In the `docker-compose.yml` file, we would reference the respective `.env` files (they would be put in `{{ lemmy_base_dir }}/inventory/host_vars/{{ domain }}/lemmyui.env,pictrs.env,postgres.env,etc`. 

This change also allows operators to create/modify docker environment variables in the future without us having to alter the playbooks each time or have to distribute and maintain `.env` sample templates. 

#### Example playbook

The variables references would be located in `vars.yml`. This is just an example of the proposed logic. 

```yaml
---
- hosts: localhost
  become: no
  vars:
    postgres_password: IAteTooManySourPatchKids
    docker_env_vars:
      - name: "pictrs"
        variables:
          PICTRS__SERVER__API_KEY: "{{ postgres_password }}"
          PICTRS__MEDIA__VIDEO_CODEC: vp9
          PICTRS__MEDIA__GIF__MAX_WIDTH: 256
          PICTRS__MEDIA__GIF__MAX_HEIGHT: 256
          PICTRS__MEDIA__GIF__MAX_AREA: 65536
          PICTRS__MEDIA__GIF__MAX_FRAME_COUNT: 400
          PICTRS_OPENTELEMETRY_URL: http://otel:4137
          RUST_LOG: debug
          RUST_BACKTRACE: full
      #   PICTRS__STORE__TYPE: object_storage 
      #   PICTRS__STORE__ENDPOINT: '<S3 endpoint>'
      #   PICTRS__STORE__BUCKET_NAME: '<bucket name>'
      #   PICTRS__STORE__REGION: '<region>'
      #   PICTRS__STORE__USE_PATH_STYLE: false 
      #   PICTRS__STORE__ACCESS_KEY: '<access key>'
      #   PICTRS__STORE__SECRET_KEY: '<secret key>'
      - name: "lemmyui"
        variables:
          LEMMY_UI_HTTPS: true
      - name: "postgres"
        variables:
          POSTGRES_USER: lemmy
          POSTGRES_PASSWORD: "{{ postgres_password }}"
          POSTGRES_DB: lemmy
          
  tasks:
    - name: Create environment variable files
      copy:
        dest: "{{ item.name }}.env"
        content: |
          {% for key, value in item.variables.items() %}
          {{ key }}={{ value }}
          {% endfor %}
      loop: "{{ docker_env_vars }}"
```

```bash
<snip>
TASK [Create environment variable files] ************************************************************
changed: [localhost] => (item={'name': 'pictrs', 'variables': {'PICTRS__SERVER__API_KEY': 'IAteTooManySourPatchKids', 'PICTRS__MEDIA__VIDEO_CODEC': 'vp9', 'PICTRS__MEDIA__GIF__MAX_WIDTH': 256, 'PICTRS__MEDIA__GIF__MAX_HEIGHT': 256, 'PICTRS__MEDIA__GIF__MAX_AREA': 65536, 'PICTRS__MEDIA__GIF__MAX_FRAME_COUNT': 400, 'PICTRS_OPENTELEMETRY_URL': 'http://otel:4137', 'RUST_LOG': 'debug', 'RUST_BACKTRACE': 'full'}})
changed: [localhost] => (item={'name': 'lemmyui', 'variables': {'LEMMY_UI_HTTPS': True}})
changed: [localhost] => (item={'name': 'postgres', 'variables': {'POSTGRES_USER': 'lemmy', 'POSTGRES_PASSWORD': 'IAteTooManySourPatchKids', 'POSTGRES_DB': 'lemmy'}})
</snip>
```

```shell
bash-3.2$ ls *.env
lemmyui.env     pictrs.env      postgres.env

bash-3.2$ cat lemmyui.env
LEMMY_UI_HTTPS=True

bash-3.2$ cat pictrs.env
PICTRS__SERVER__API_KEY=IAteTooManySourPatchKids
PICTRS__MEDIA__VIDEO_CODEC=vp9
PICTRS__MEDIA__GIF__MAX_WIDTH=256
PICTRS__MEDIA__GIF__MAX_HEIGHT=256
PICTRS__MEDIA__GIF__MAX_AREA=65536
PICTRS__MEDIA__GIF__MAX_FRAME_COUNT=400
PICTRS_OPENTELEMETRY_URL=http://otel:4137
RUST_LOG=debug
RUST_BACKTRACE=full

bash-3.2$ cat postgres.env
POSTGRES_USER=lemmy
POSTGRES_PASSWORD=IAteTooManySourPatchKids
POSTGRES_DB=lemmy
```

If you and @dessalines  prefer this method, I'll go ahead and get everything done and merged in :).

codyro: I made this change in another branch/PR in my repo so you can see what it would look like. It's ready to merge and has been tested: https://github.com/codyro/lemmy-ansible/pull/3/files

dessalines: I personally prefer this PR to your new one that creates additional `.env` files. 

Mainly because the latter *duplicates* settings, so it might not be clear that when you edit one of those `.env` files, and run the playbook, it will actually just regenerate them from the `vars.yml` file. So they are really useless files IMO.

codyro: I'm in agreement, although I don't have strong feelings either way. 

I'll let you and @Nutomic decide the approach you guys would prefer to merge.

I'm also open to other ideas if anyone has any!

Nutomic: Having a separate .env file for each service seems totally unnecessary. All the ones I see already have prefixes PICTRS, POSTGRES etc so there can be no confusion. In case of RUST_LOG, you can define LEMMY_RUST_LOG in .env and then rename it like this:
```
environment:
  RUST_LOG: ${LEMMY_RUST_LOG}
```

dessalines: IMO the less places we have to configure, the better, so I'd much prefer this single `vars.yml` you've done here.

examples/vars.yml
```diff
@@ -0,0 +1,25 @@
...
+pictrs_env_vars:
+  - PICTRS__SERVER__API_KEY: "{{ postgres_password }}"
+  - PICTRS__MEDIA__VIDEO_CODEC: vp9
+  - PICTRS__MEDIA__GIF__MAX_WIDTH: 256
+  - PICTRS__MEDIA__GIF__MAX_HEIGHT: 256
+  - PICTRS__MEDIA__GIF__MAX_AREA: 65536
+  - PICTRS__MEDIA__GIF__MAX_FRAME_COUNT: 400
+  - PICTRS_OPENTELEMETRY_URL: http://otel:4137
+  - RUST_LOG: debug
+  - RUST_BACKTRACE: full
```
ticoombs, line 14: Should we really have the open telemetry, debug mode, and full backtraces enabled for production? 

I think we should make these 3 commented as well

examples/vars.yml
ticoombs: > I'd much prefer this single `vars.yml`

@codyro I agree. Singular variable file per "host" is the simplest approach that will cause the least amount of issues.

Having them inside the playbook is a non-starter because the whole point of moving the variables out is so that users of the playbooks do not have to edit playbooks. The goal is for them to make changes outside the playbooks (env_file) and then run the playbook.

The lack of defaults is the only thing that annoys me. ie; what happens if a user were to forget this crucial step of copying the variables.

examples/hosts
```diff
@@ -1,14 +1,15 @@
...
+# - `ansible_user=root`: replace `root` with your the username you use to connect to your ssh server
 # - `domain=example.com`: replace `example.com` with your lemmy domain
 # - `letsencrypt_contact_email=your@email.com` replace `your@email.com` with your email address,
 #                                              to get notifications if your ssl cert expires
 # - `lemmy_base_dir=/srv/lemmy`: the location on the server where lemmy can be installed, can be any folder
 #                                if you are upgrading from a previous version, set this to `/lemmy`
 # - `lemmy_version`: <Optional> The back end version.
 # - `lemmy_ui_version`: <Optional> overrides the front end version.
-myuser@example.com  domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy
+example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy
```
ticoombs, line 12: where `{{ domain }}` would be the first `example.com`

But this can be anything if you have your own `.ssh/config`

lemmy.yml
```diff
@@ -166,9 +166,6 @@
             owner: "1000"
             group: "1000"
 
-      vars:
```
ticoombs, line 169: i'm pretty sure `host_vars` are mapped against the host in the `inventory/hosts` file. As such, in the current playbooks we never import the whole file. We use `host_vars` but not how it was originally intended via ansible. 

Pretty sure we still need to actually import the variable file `inventory/host_vars/{{ domain }}/vars.yml` 

Because otherwise we make assumptions that the users `hosts` file exactly matches our format (see other comment)

ticoombs (changes requested):

examples/hosts
codyro: I think it serves its purpose as an example.

lemmy.yml
codyro: I am not following your comment. Assuming everyone follows the semantics of the README/docs, Ansible will auto-load the `vars.yml` in my example. You do not need to explicitly import it.

examples/vars.yml
codyro: Can you submit a new issue for this? I can see an argument for both:

* Keeping it enabled is excessive and noisy/unnecessary

* Enabling it by default gives useful debug for users/developers who may not be savvy enough to know how to adjust this environment to get useful debug output without some finagling

lemmy.yml
ticoombs: > You do not need to import it explicitly

If your testing what does your host file look like. Does your "server" match your `domain=<server>`

Testing:  
- inventory/hosts
```
[lemmy]
my-server-name-which-doesnt-match-domain  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy
```

- inventory/host_vars/example.com/vars.yml
```
postgres_env_vars:
  - POSTGRES_USER: lemmy
```

- test.yml
```
--
- hosts: lemmy

  tasks:
  - name: Display all variables/facts known for a host
    ansible.builtin.debug:
      msg:
        - "Provisioning based on postgres_env_vars which is: {{ lookup('ansible.builtin.env', 'postgres_env_vars') }}"
```

result:
```
ok: [my-server-name-which-doesnt-match-domain] => {
    "msg": [
        "Provisioning based on pictrs_vars which is: "
    ]
}
```

----
changing: inventory/hosts so that `example.com` is now the "host"
```
[lemmy]
example.com ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy
```

Results in ansible actually merging the host_vars. That is why they are called host_vars. As they are specific for the host.

Proof:

```
PLAY [lemmy] ********************************************************************************************************
ERROR! failed to combine variables, expected dicts but got a 'dict' and a 'AnsibleUnicode':
ansible error continues because we are attempting to print a dictionary object
```

PR #161, opened 2023-09-02
Fix incorrect/inconsistent usage of `proxy_set_header` 

### LINKED ISSUE #155: Internal nginx not sending headers to backend on GET
[Per the nginx documentation](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header), proxy_set_header directives are only inherited from the previous config level if none are specified at the current level. 

So in this snippet from the nginx_internal.conf template, X-Real-IP, Host, and X-Forwarded-For are set at the server{ } level and are inherited by some of the location blocks and not others. Specifically, the backend redirect at the bottom contains other proxy_set_header directives, that will inhibit the inheritance.

```
    server {

        <*snip*>

        # Send actual client IP upstream
        proxy_set_header Host $host;

        # frontend general requests
        location / {
            proxy_pass $proxpass;
            rewrite ^(.+)/+$ $1 permanent;
        }

        # security.txt
        location = /.well-known/security.txt {
            proxy_pass "http://lemmy-ui";
        }

        # backend
        location ~ ^/(api|pictrs|feeds|nodeinfo|.well-known) {
            proxy_pass "http://lemmy";

            # proxy common stuff
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "upgrade";
        }
    }
```

codyro:
Fixes the issue described in #155.

Tested against the normal spread of targets (Debian 11,12,Ubuntu 22.04 LTS, AL9/RHEL9). 

- Distribute proxy_params to nginx VM, include it as necessary in config
- Resolves #155

### DIFF
--- a/templates/nginx_internal.conf
+++ b/templates/nginx_internal.conf
@@ -59,9 +59,7 @@ http {
         client_max_body_size 20M;
 
         # Send actual client IP upstream
-        proxy_set_header X-Real-IP $remote_addr;
-        proxy_set_header Host $host;
-        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        include proxy_params;
 
         # frontend general requests
         location / {
@@ -79,6 +77,7 @@ http {
             proxy_pass "http://lemmy";
 
             # proxy common stuff
+            include proxy_params;
             proxy_http_version 1.1;
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection "upgrade";
--- a/templates/nginx_internal.conf
+++ b/templates/nginx_internal.conf
@@ -76,11 +76,8 @@ http {
         location ~ ^/(api|pictrs|feeds|nodeinfo|.well-known) {
             proxy_pass "http://lemmy";
 
-            # proxy common stuff
+            # Send actual client IP upstream
             include proxy_params;
-            proxy_http_version 1.1;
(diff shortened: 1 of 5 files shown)


templates/nginx_internal.conf
```diff
@@ -79,6 +77,7 @@ http {
             proxy_pass "http://lemmy";
 
             # proxy common stuff
+            include proxy_params;
             proxy_http_version 1.1;
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection "upgrade";
```
Nutomic, line 83: I believe these lines were only needed for websocket. Now that websocket is gone it should be fine to remove them.

codyro: Just to clarify, the legacy lines are the `proxy_set_header Upgrade $http_upgrade;` and `proxy_set_header Connection "upgrade";`, not the proxy related ones?

Once confirmed, I'll get this fixed up.

dessalines: That's correct, it looks like its those two "upgrade" lines: https://nginx.org/en/docs/http/websocket.html

codyro: I removed the `proxy_http_version 1.1;` section as well, as I believe that was a requirement of the websockets. Please correct me if I'm wrong.

dessalines (approved):

PR #165, opened 2023-09-06
Fix nginx connect 111 error.

dessalines:
On lemmy.ml , we've been periodically getting some unrecoverable errors with our internal nginx proxy. Only doing `sudo docker compose restart proxy` fixes it for a few hours.

Error: `connect() failed (111: Connection refused)`

Context and potential fix: https://stackoverflow.com/questions/51009692/docker-nginx-proxy-nginx-connect-failed-111-connection-refused-while-connec

I'm seeing if this will fix the issue now.

### DIFF
--- a/templates/nginx_internal.conf
+++ b/templates/nginx_internal.conf
@@ -21,29 +21,19 @@ http {
     # Learn more about nginx maps here http://nginx.org/en/docs/http/ngx_http_map_module.html
     map "$request_method:$http_accept" $proxpass {
         # If no explicit matches exists below, send traffic to lemmy-ui
-        default "http://lemmy-ui";
+        default "http://lemmy-ui:1234";
 
         # GET/HEAD requests that accepts ActivityPub or Linked Data JSON should go to lemmy.
         #
         # These requests are used by Mastodon and other fediverse instances to look up profile information,
         # discover site information and so on.
-        "~^(?:GET|HEAD):.*?application\/(?:activity|ld)\+json" "http://lemmy";
+        "~^(?:GET|HEAD):.*?application\/(?:activity|ld)\+json" "http://lemmy:8536";
 
         # All non-GET/HEAD requests should go to lemmy
         #
         # Rather than calling out POST, PUT, DELETE, PATCH, CONNECT and all the verbs manually
         # we simply negate the GET|HEAD pattern from above and accept all possibly $http_accept values
-        "~^(?!(GET|HEAD)).*:" "http://lemmy";
-    }
-
-    upstream lemmy {
-        # this needs to map to the lemmy (server) docker service hostname
-        server "lemmy:8536";
-    }
-
-    upstream lemmy-ui {
-        # this needs to map to the lemmy-ui docker service hostname
-        server "lemmy-ui:1234";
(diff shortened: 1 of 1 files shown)


codyro (approved):
Looks much cleaner. Tested and ran through `gixy` for kicks.

Was the issue you were experiencing being caused by nginx and the `upstream` directive?

dessalines:
Yes I believe so. I'm going to leave it running on lemmy.ml for a few days yet, to make sure that it fixes the issue.

templates/nginx_internal.conf
```diff
@@ -21,29 +21,19 @@ http {
     # Learn more about nginx maps here http://nginx.org/en/docs/http/ngx_http_map_module.html
     map "$request_method:$http_accept" $proxpass {
         # If no explicit matches exists below, send traffic to lemmy-ui
-        default "http://lemmy-ui";
+        default "http://lemmy-ui:1234";
```
Nutomic, line 24: So these lines were all missing the port? Not sure why it worked before. 

Maybe we should change all the docker containers to listen on standard http port (80), would make things easier than these random port numbers.

codyro:
> EDIT: tried out gixy, seems like it passes even when you add errors... might have to do with our template vars being not readable.

It doesn't like the template variables, but it's looking for common misconfigurations that can lead to some security issues/information disclosures. 

Adding `nginx -t` to the CI could be helpful to ensure the configuration doesn't become invalid with a change. You would want to have Ansible process the template first, then run `nginx -t` and optionally `gixy`.

dessalines:
There might be an ansible job that can check that. Not sure if it's possible to get that to run in the CI tho.

dessalines:
Yeah, if possible. If not, maybe it might be possible to perform some kind of `dry-run` ansible check, but I'm not sure.

ticoombs:
Ports are redundant in when used with upstreams. the `http://lemmy` here is pointing to `http://<upstream>`
And in the upstream you have the ports. (But you've gotten rid of them)

Restarting the nginx container screams DNS issues specifically it happens when your `lemmy` containers restart on their own. See: https://stackoverflow.com/questions/35744650/docker-network-nginx-resolver for a little bit more info regarding the solution.

Basically; nginx upon start resolves `lemmy` via docker's resolver which gets all ips of your `lemmy` containers. (i have 5 for example) the TTL by default is super high (1 hour, i cant remember) especially if you only have 1 lemmy container and it kills/restarts itself _and_ gets a new IP.

Try adding to your `http` block to the nginx_internal: 
```
   resolver 127.0.0.11 ipv6=off valid=5s;
```

This forces all dns resolution to be refreshed every ~5 seconds. So then atleast if you scale-up/down the nginx container should no longer need to be restarted.

There are assumptions I'm making about your infra, and the actual problems relating to "connect 111 error"

codyro:
Is there anyway I can replicate this in a testing environment? You've piqued my interest. If I just lambast the instance with traffic does it happen? A specific API endpoint perhaps? Or is it difficult to reproduce?

ticoombs (approved):

dessalines:
I'm not sure of a way to test it, just one of those strange docker networking issues.

PR #166, opened 2023-09-10
Fix ansible-lint warnings

Nothing4You:
- add ansible-lint check to CI
- reformat playbooks to conform to strict ansible-lint checks
- set explicit modes for all files and directories created in playbook
- only ensure presence, not latest for packages installed by the playbook

### `lemmy-almalinux.yml`

- `{{ lemmy_base_dir }}/{{ domain }}` and directories within will now be created with `chmod 0755`.
This was previously left to the system's default umask, which typically results in `0755` for directories.
Related linter rule: [risky-file-permissions](https://ansible.readthedocs.io/projects/lint/rules/risky-file-permissions/)

### `lemmy.yml`

- Docker GPG key will now be downloaded with `chmod 0644`.
This was previously left to the system's default umask, which typically results in `0644` for files.
Related linter rule: [risky-file-permissions](https://ansible.readthedocs.io/projects/lint/rules/risky-file-permissions/)
- Package installations will now only ensure presence of packages instead of ensuring the latest version of the package is installed.
I'm not sure which is the desired behavior here, I'd expect to have a separate dedicated "full update" task if it's desired to also perform a system update at this stage.
For example, when updating docker to a newer version, this will also restart the docker service and in turn all containers.
Related linter rule: [package-latest](https://ansible.readthedocs.io/projects/lint/rules/package-latest/)

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -10,3 +10,8 @@ pipeline:
       - ansible-playbook lemmy.yml --syntax-check
       - ansible-playbook lemmy-almalinux.yml --syntax-check
       - ansible-playbook uninstall.yml --syntax-check
+  ansible_lint:
+    image: alpine:3
+    commands:
+      - apk add ansible ansible-lint
+      - ansible-lint --warn-list experimental lemmy.yml lemmy-almalinux.yml uninstall.yml
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -1,16 +1,18 @@
 ---
-- hosts: all
+- name: Install Lemmy
+  hosts: all
   gather_facts: true
   pre_tasks:
     # This is not needed for this playbook as it predates its existence
     # But we're keeping it for funsies :)
-    - name: check lemmy_base_dir
+    - name: Check lemmy_base_dir
       ansible.builtin.fail:
         msg: "`lemmy_base_dir` is unset. if you are upgrading from an older version, add `lemmy_base_dir=/lemmy` to your inventory file."
       when: lemmy_base_dir is not defined
   handlers:
-    - name: reload nginx
+    - name: Reload nginx
       ansible.builtin.command: nginx -s reload
+      changed_when: true
   vars:
     lemmy_port: "{{ 32767 | random(start=1024) }}"
   tasks:
@@ -30,7 +32,7 @@
         mirrorlist: https://mirrors.almalinux.org/mirrorlist/$releasever/crb
         gpgcheck: true
         gpgkey: file:///etc/pki/rpm-gpg/RPM-GPG-KEY-AlmaLinux-9
-        enabled: yes
+        enabled: true
       tags:
         - dependencies
 
@@ -72,12 +74,14 @@
         - http
         - https
       when: "'firewalld.service' in ansible_facts.services and ansible_facts.services['firewalld.service'].state == 'running'"
+      changed_when: true
       tags:
         - firewalld
 
     - name: Reload firewalld
       ansible.builtin.command: firewall-cmd --reload
       when: "'firewalld.service' in ansible_facts.services and ansible_facts.services['firewalld.service'].state == 'running'"
+      changed_when: true
       tags:
         - firewalld
 
@@ -96,6 +100,7 @@
         path: "{{ item.path }}"
         owner: "{{ item.owner }}"
         state: directory
+        mode: "0755"
       loop:
         - path: "{{ lemmy_base_dir }}/{{ domain }}/"
           owner: "root"
@@ -106,7 +111,7 @@
 
     - name: Set lemmy_port fact
       ansible.builtin.set_fact:
-        lemmy_port: "{{ 32767 |random(start=1024) }}"
+        lemmy_port: "{{ 32767 | random(start=1024) }}"
       tags:
         - always
 
@@ -123,8 +128,8 @@
           dest: "{{ lemmy_base_dir }}/{{ domain }}/nginx_internal.conf"
           mode: "0644"
       vars:
-        lemmy_docker_image: "docker.io/dessalines/lemmy:{{ lemmy_version | default( lookup('file', 'VERSION') )}}"
-        lemmy_docker_ui_image: "docker.io/dessalines/lemmy-ui:{{ lemmy_ui_version | default(lemmy_version | default(lookup('file', 'VERSION')))}}"
+        lemmy_docker_image: "docker.io/dessalines/lemmy:{{ lemmy_version | default(lookup('file', 'VERSION')) }}"
+        lemmy_docker_ui_image: "docker.io/dessalines/lemmy-ui:{{ lemmy_ui_version | default(lemmy_version | default(lookup('file', 'VERSION'))) }}"
       tags:
         - docker
         - podman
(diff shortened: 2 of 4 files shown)


Nothing4You:
there appears to be an issue with CI still, need to look into that

Nothing4You:
the warnings don't seem to be causing problems.
running this locally on macOS with latest ansible-lint and ansible versions it doesn't print any warnings, but running the older ansible-lint version on an alpine 3.18 container does reproduce this.
running this on alpine edge with

> ansible-lint 6.19.0 using ansible-core:2.15.3 ansible-compat:4.1.10 ruamel-yaml:0.17.28 ruamel-yaml-clib:0.2.7

does not have these warnings anymore.

the actual linter warnings show up in both alpine 3.18 and alpine edge, so I don't think this has practical impact and should go away when alpine 3.19 is released in november.

codyro:
Thank you for this--the playbooks need to be cleaned up, and this goes a long way! My only concern with this is if `prettier` doesn't agree with `ansible-lint` about formatting, causing the CI changes to conflict.

Perhaps we can consider running `ansible-lint` without `--strict`?

.woodpecker.yml
```diff
@@ -6,7 +6,9 @@ pipeline:
   check_ansible_format:
     image: alpine:3
     commands:
-      - apk add ansible
+      - apk add ansible ansible-lint
       - ansible-playbook lemmy.yml --syntax-check
       - ansible-playbook lemmy-almalinux.yml --syntax-check
       - ansible-playbook uninstall.yml --syntax-check
+      - ansible-lint --version
+      - ansible-lint --strict lemmy.yml lemmy-almalinux.yml uninstall.yml
```
codyro, line 14: I'd consider not using `--strict` to avoid CI conflicts with `prettier` (which is *very* opinionated). 

Is there any reason for the `ansible-lint --version` here, or is it just a leftover debug line?

Nothing4You: `ansible-lint --version` was added for my CI troubleshooting, removed now.

do you know of any conflicts with `prettier`?

I've included strict to ensure warnings like [`jinja[spacing]`](https://ansible.readthedocs.io/projects/lint/rules/jinja/) would be addressed/flagged:

```
jinja[spacing]: Jinja2 spacing could be improved: {{ 32767 |random(start=1024) }} -> {{ 32767 | random(start=1024) }} (warning)
lemmy-almalinux.yml:110 Jinja2 template rewrite recommendation: `{{ 32767 | random(start=1024) }}`.

jinja[spacing]: Jinja2 spacing could be improved: docker.io/dessalines/lemmy-ui:{{ lemmy_ui_version | default(lemmy_version | default(lookup('file', 'VERSION')))}} -> docker.io/dessalines/lemmy-ui:{{ lemmy_ui_version | default(lemmy_version | default(lookup('file', 'VERSION'))) }} (warning)
lemmy-almalinux.yml:127 Jinja2 template rewrite recommendation: `docker.io/dessalines/lemmy-ui:{{ lemmy_ui_version | default(lemmy_version | default(lookup('file', 'VERSION'))) }}`.

jinja[spacing]: Jinja2 spacing could be improved: docker.io/dessalines/lemmy:{{ lemmy_version | default( lookup('file', 'VERSION') )}} -> docker.io/dessalines/lemmy:{{ lemmy_version | default(lookup('file', 'VERSION')) }} (warning)
lemmy-almalinux.yml:127 Jinja2 template rewrite recommendation: `docker.io/dessalines/lemmy:{{ lemmy_version | default(lookup('file', 'VERSION')) }}`.

[…]

                        Rule Violation Summary
 count tag                      profile    rule associated tags
[…]
    37 jinja[spacing]           basic      formatting (warning)
```

If necessary due to `prettier` conflicts, individual linter checks could still be ignored.

codyro: I am unaware of any conflicts off-hand. However, it's almost guaranteed to happen.

`prettier` does a good job explaining what it is trying to achieve  [here](https://prettier.io/docs/en/comparison.html) and why we shouldn't mix it with `--strict`. 

> In other words, use Prettier for formatting and linters for catching bugs!

Nothing4You: `prettier` isn't aware of ansible-specific formatting, such as jinja templated strings.
With the current yaml, `jinja[spacing]` was the only type of warning emitted, and imo it's reasonable for this to fail CI.
As of right now, on the latest version of `ansible-lint`, the following rules only emit warnings:

> Current version default value is: experimental, jinja[spacing], fqcn[deep]

I don't see how this could in any way conflict with `prettier`.

While it probably makes sense to ignore `experimental` checks, and `fqcn[deep]` checks are not relevant to this repo at all (currently), `jinja[spacing]` is relevant.

A better alternative seems to be using `--warn-list experimental`, which I've updated this PR to.

codyro: I just took a closer look, and you're right. I don't think there will be any conflicts (well, any that would cause a loop with the CI). 

I think `--warn-list experimental` is a good alternative.

codyro (approved):

ticoombs (approved):
Looks good. Thanks for the linting fixes!

Nothing4You:
rebased to resolve conflicts

.woodpecker.yml
```diff
@@ -6,7 +6,8 @@ pipeline:
   check_ansible_format:
     image: alpine:3
     commands:
-      - apk add ansible
+      - apk add ansible ansible-lint
       - ansible-playbook lemmy.yml --syntax-check
       - ansible-playbook lemmy-almalinux.yml --syntax-check
       - ansible-playbook uninstall.yml --syntax-check
+      - ansible-lint --warn-list experimental lemmy.yml lemmy-almalinux.yml uninstall.yml
```
Nutomic, line 13: You can consider making separate steps for these so its easier to tell whats failing. But like this is also okay.

Nothing4You: Not directly related to this PR, but have you considered using GitHub actions?
`ansible-lint` can produce [SARIF](https://sarifweb.azurewebsites.net/) output, which [GitHub can use for annotations directly in the diff](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning).
As far as I can tell, third party CIs are not able to do this.

Nothing4You: I've moved it to a separate woodpecker pipeline now.

Nutomic: We want to avoid relying on another proprietary service. Woodpecker is selfhosted and fully open source so its a better match for Lemmy.

codyro:
All of these items need to be addressed. However, they're not related to this PR. Please submit a new issue/PR so we can continue cleaning this project up!

sanatsathaye:
> All of these items need to be addressed. However, they're not related to this PR. Please submit a new issue/PR so we can continue cleaning this project up!

I mean, points 1 and 3 are directly related to the PR, if the point of the PR is to fix linter warnings (`no-changed-when` and `no-relative-paths`) then why not actually fix them rather than using a temporary workaround? :)

Anyway, considering this has already been been merged I'll make a PR when I'm able to.

Nothing4You:
Indeed, I've tried to keep everything as much as-is, as far as the logic/tasks are concerned, to focus on one improvement at a time.
This usually also simplifies reviews of changes, as smaller diffs are easier to review at a time compared to a giant diff.
This PR already includes a very large diff, but the majority of it is more or less just formatting/more explicit syntax.
Adding more changes to this PR would quickly make it a lot harder to review the diff.

sanatsathaye:
Np, that's understandable.

@codyro - Sure, think I should be able to make a PR tomorrow.

PR #167, opened 2023-09-12
Fix change that broke notify/reload, update nginx handler

codyro:
- Update handler in `lemmy-almalinux.yml` to use ansible.builtin.systemd module
- Fixes a bug introduced in #166 
- Utilize handler & systemd module for `nginx` reloading in `lemmy.yml`

I will make similar changes to the primary playbook later, but since this is breaking, I want to merge it ASAP. This is another good use case for getting proper testing (e.g., Molecule). The only hindrance I've had in that realm is requiring Docker In Docker to fully test the playbooks.

### DIFF
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -11,8 +11,9 @@
       when: lemmy_base_dir is not defined
   handlers:
     - name: Reload nginx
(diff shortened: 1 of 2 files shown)


dessalines (commented):
Thx!

dessalines (approved):

Issue #170, opened 2023-09-13
Update README `Upgrading` section, add check?


codyro:
Improve the `Upgrading` section of the README to reflect the changes/addition of the `vars.yml` file & renaming of the `passwords/postgres` file to `passwords/postgres.psk` file for users upgrading from versions that do not have this file. 

In addition to this (or alternatively), we could have the playbooks do a sanity check and notify the user if it's missing in the `pre_tasks` or elsewhere.

Refs
* https://github.com/LemmyNet/lemmy-ansible/issues/169#issuecomment-1718288517

ticoombs:
FWIW, I think the pre_tasks checks should exist, and pointing people to the readme's especially regards to the (new?) upgrade procedure.

codyro:
Done in #172

PR #172, opened 2023-09-14
Update README.md, test some pre_task checks, explicitly import vars.yml

codyro:
PoC for https://github.com/LemmyNet/lemmy-ansible/issues/170


dessalines (commented):
Seems like a good approach to me.

ticoombs (approved):

PR #175, opened 2023-09-23
feat: pictrs-safety

db0:
Adds inventory variable to enable picts-safety

pictrs gets an extra conditional for a new variable `PICTRS__MEDIA__EXTERNAL_VALIDATION` to connect the validation to pictrs-safety. 

I used `{{ domain }} for it, as I don't know how otherwise to tell pictrs which IP to connect. Feel free to adjust to use some internal IP. My docker knowledge is very superficial.

Likewise, the pictrs-safety port is supposed to be exposed to the internet to allow workers to connect to it. 

Currently I'm using the `:main` version as I'm having trouble pushing versioned docker images to ghcr.io. I'm trying to figure that out.

### DIFF
--- a/examples/hosts
+++ b/examples/hosts
@@ -9,7 +9,8 @@
 #                                if you are upgrading from a previous version, set this to `/lemmy`
 # - `lemmy_version`: <Optional> The back end version.
 # - `lemmy_ui_version`: <Optional> overrides the front end version.
-example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy
+# - `use_pictrs_safety`: <Optional> If true, a docker container for pictrs-safety will be deployed and pict-rs will be configured to validate images through it. You will also need to set up a fedi-safety worker to validate the images.
+example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy pictrs_safety=false
 
 [all:vars]
 ansible_connection=ssh
--- a/examples/hosts
+++ b/examples/hosts
@@ -9,7 +9,7 @@
 #                                if you are upgrading from a previous version, set this to `/lemmy`
 # - `lemmy_version`: <Optional> The back end version.
 # - `lemmy_ui_version`: <Optional> overrides the front end version.
-# - `use_pictrs_safety`: <Optional> If true, a docker container for pictrs-safety will be deployed and pict-rs will be configured to validate images through it. You will also need to set up a fedi-safety worker to validate the images.
+# - `pictrs_safety`: <Optional> If true, a docker container for pictrs-safety will be deployed and pict-rs will be configured to validate images through it. You will also need to set up a fedi-safety worker to validate the images.
 example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy pictrs_safety=false
 
 [all:vars]
--- a/examples/vars.yml
+++ b/examples/vars.yml
@@ -35,3 +35,12 @@ lemmyui_env_vars:
 
 postfix_env_vars:
   - POSTFIX_myhostname: "{{ domain }}"
+
+pictrs_safety_env_vars:
+  # Use this in your fedi-safety to allow your worker to authenticate to pictrs-safety
+  - FEDIVERSE_SAFETY_WORKER_AUTH: "{{ lookup('password', 'inventory/host_vars/{{domain}}/passwords/pictrs_safety_worker_auth.psk chars=ascii_letters,digits length=15') }}"
+  - FEDIVERSE_SAFETY_IMGDIR: "/tmp/images"
+  - USE_SQLITE: 1
+  - secret_key: "{{ lookup('password', 'inventory/host_vars/{{domain}}/passwords/pictrs_safety_secret.psk chars=ascii_letters,digits length=80') }}"
+  - SCAN_BYPASS_THRESHOLD: 10
+  - MISSING_WORKER_THRESHOLD: 5
--- a/examples/vars.yml
+++ b/examples/vars.yml
@@ -1,4 +1,7 @@
-postgres_password: "{{ lookup('password', 'inventory/host_vars/{{domain}}/passwords/postgres.psk chars=ascii_letters,digits') }}"
+postgres_password: "{{ lookup('password', 'inventory/host_vars/{{ domain }}/passwords/postgres.psk chars=ascii_letters,digits') }}"
+# Next two only relevant if pictrs_safety == True
+pictrs_safety_worker_auth: "{{ lookup('password', 'inventory/host_vars/{{ domain }}/passwords/pictrs_safety_worker_auth.psk chars=ascii_letters,digits length=15') }}"
+pictrs_safety_secret: "{{ lookup('password', 'inventory/host_vars/{{ domain }}/passwords/pictrs_safety_secret.psk chars=ascii_letters,digits length=80') }}"
 
 # You can set any pict-rs environmental variables here. They will populate the templates/docker-compose.yml file.
 # https://git.asonix.dog/asonix/pict-rs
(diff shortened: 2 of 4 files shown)


templates/docker-compose.yml
```diff
@@ -72,6 +72,9 @@ services:
       - {{ key }}={{ value }}
 {% endfor %}
 {% endfor %}
+{% endif %}
+{% if pictrs_safety is defined and pictrs_safety|bool == true %}
+      - PICTRS__MEDIA__EXTERNAL_VALIDATION=http://{{ domain }}:14051/api/v1/scan/IPADDR
```
codyro, line 77: It is worth considering proxying this via nginx instead of directly. This has the added benefit of being able to use `https` easily.

templates/docker-compose.yml
```diff
@@ -112,3 +115,29 @@ services:
 {% endif %}
     restart: "always"
     logging: *default-logging
+
+{% if pictrs_safety is defined and pictrs_safety|bool == true %}
+  pictrs-safety:
+    image: ghcr.io/db0/pictrs-safety:main
```
codyro, line 121: You already made note of this, but we should ensure this us a tagged version.

codyro (commented):
I'll give it a more in-depth look later :). 

Thanks for the PR!

examples/hosts
```diff
@@ -9,7 +9,8 @@
 #                                if you are upgrading from a previous version, set this to `/lemmy`
 # - `lemmy_version`: <Optional> The back end version.
 # - `lemmy_ui_version`: <Optional> overrides the front end version.
-example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy
+# - `use_pictrs_safety`: <Optional> If true, a docker container for pictrs-safety will be deployed and pict-rs will be configured to validate images through it. You will also need to set up a fedi-safety worker to validate the images.
+example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy pictrs_safety=false
```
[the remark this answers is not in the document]
db0: fixed

templates/docker-compose.yml
Nutomic: There is no reason to open a public port and go through the public domain for this. It should work just fine to use the internal Docker network, by setting the url `http://pictrs-safety:14051/api/v1/scan/IPADDR`. Then the port below doesnt need to be exposed.

Another matter is that some admins may want to run pictrs-safety on a separate server with GPU, or run an entirely different verification software. With that in mind I would consider only adding a new parameter `pictrs_validation_url` and linking instructions to install pictrs-safety manually.

db0: About the ports, if pictrs-safety is installed, you likely **do** want the external connection to allow GPU workers running fedi-safety to connect and pick up validation jobs.

Nutomic: Ah so the pictrs-safety image below doesnt do the validation on its own, but only hands out validation jobs to external workers? Then this setup makes sense. Anyway you should also include a documentation link which explains how to set this all up. You could write a page about this in the lemmy docs (or somewhere else).

db0: the pictrs-safety repo readme does provide full documentation, including a docker-compose sample
The fedi-safety likewise provides full documentation for connecting to pictrs-safety and also a docker container if someone wants to use it that way.

If this is accepted into the ansible repo, I'll work on adding doc to lemmy docs as well

templates/docker-compose.yml
```diff
@@ -112,3 +115,29 @@ services:
 {% endif %}
     restart: "always"
     logging: *default-logging
+
+{% if pictrs_safety is defined and pictrs_safety|bool == true %}
+  pictrs-safety:
+    image: ghcr.io/db0/pictrs-safety:v1.2.2
+    hostname: "pictrs-safety"
+    networks:
```
dessalines, line 123: These networks aren't used anymore.

examples/hosts
```diff
@@ -9,7 +9,8 @@
 #                                if you are upgrading from a previous version, set this to `/lemmy`
 # - `lemmy_version`: <Optional> The back end version.
 # - `lemmy_ui_version`: <Optional> overrides the front end version.
-example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy
+# - `pictrs_safety`: <Optional> If true, a docker container for pictrs-safety will be deployed and pict-rs will be configured to validate images through it. You will also need to set up a fedi-safety worker to validate the images.
+example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy pictrs_safety=false
```
dessalines, line 13: Just to verify, this new var can be missing and the playbook will still run, correct? I see an `isDefined` check below so that's probably the case.

db0: Correct

templates/docker-compose.yml
db0: Oh, should I remove them, or replace them with something? As I said, I'm not that good with docker

codyro: You should be able to remove them. They will cause an error as-is.

examples/vars.yml
```diff
@@ -35,3 +35,12 @@ lemmyui_env_vars:
 
 postfix_env_vars:
   - POSTFIX_myhostname: "{{ domain }}"
+
+pictrs_safety_env_vars:
+  # Use this in your fedi-safety to allow your worker to authenticate to pictrs-safety
+  - FEDIVERSE_SAFETY_WORKER_AUTH: "{{ lookup('password', 'inventory/host_vars/{{domain}}/passwords/pictrs_safety_worker_auth.psk chars=ascii_letters,digits length=15') }}"
```
ticoombs, line 41: Please move this to the top of the file like all the other passwords we have. Assign it to a variable, and then have it referenced here.

Also linting wise, we need spaces around domain: `{{ domain }}`.

(If you run ansible-lint against the file it will tell you all the errors)

examples/vars.yml
```diff
@@ -35,3 +35,12 @@ lemmyui_env_vars:
...
+  - FEDIVERSE_SAFETY_IMGDIR: "/tmp/images"
+  - USE_SQLITE: 1
```
ticoombs, line 43: Shouldn't this be TRUE or FALSE? A `1` indicates something different that could be misconstrued.

examples/vars.yml
```diff
@@ -35,3 +35,12 @@ lemmyui_env_vars:
...
+  - secret_key: "{{ lookup('password', 'inventory/host_vars/{{domain}}/passwords/pictrs_safety_secret.psk chars=ascii_letters,digits length=80') }}"
```
ticoombs, line 44: Linting: 
- needs to be `{{ domain }}`
- please use a better name for the variable and ensure it is capitalized: FEDIVERSE_SAFETY_SECRET_KEY

Also see other comment regarding moving to the top.

templates/docker-compose.yml
```diff
@@ -112,3 +115,29 @@ services:
 {% endif %}
     restart: "always"
     logging: *default-logging
+
+{% if pictrs_safety is defined and pictrs_safety|bool == true %}
+  pictrs-safety:
+    image: ghcr.io/db0/pictrs-safety:v1.2.2
+    hostname: "pictrs-safety"
```
ticoombs, line 122: you don't need quotes here

templates/docker-compose.yml
```diff
@@ -112,3 +115,29 @@ services:
...
+{% for item in pictrs_safety_env_vars %}
+{% for key, value in item.items() %}
+      - {{ key }}={{ value }}
+{% endfor %}
+{% endfor %}
+{% endif %}
+    ports:
+      - "14051:14051"
+    expose:
+      - '14051'
```
ticoombs, line 137: `expose` is redundant iirc as `ports` takes precedence over expose.

ticoombs (changes requested):

examples/vars.yml
db0: The `secret_key` env variable is the one used in the code. I can't just rename it

templates/docker-compose.yml
db0: They don't hurt though. If you insist on the aesthetics I can remove them

examples/vars.yml
db0: This is how it's parsed in the code and I don't like to be converting string-booleans to booleans in python. It's simpler to compare the ints.

ticoombs (approved):

ticoombs:
From a code standpoint it looks fine now.

dessalines (approved):
@codyro once it looks good to you, feel free to merge.

codyro:
Everything looks good/tested on the normal targets.

At some point, we should consider showing an example of using SSL/TLS so the pictrs-safety<->worker data can be encrypted (image body is sent).

PR #176, opened 2023-09-27
Add vars.yml to woodpecker pipeline

codyro:
Run `ansible-lint` and `prettier` against `examples/vars.yml`. 

We could just run `prettier` on all `*.yml` if we wanted to instead.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -2,7 +2,7 @@ pipeline:
   prettier_markdown_check:
     image: tmknom/prettier
     commands:
-      - prettier -c "*.md" "*.yml"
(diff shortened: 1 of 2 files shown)


.woodpecker.yml
```diff
@@ -2,7 +2,7 @@ pipeline:
   prettier_markdown_check:
     image: tmknom/prettier
     commands:
-      - prettier -c "*.md" "*.yml"
+      - prettier -c "*.md" "*.yml" "examples/vars.yml"
```
dessalines, line 5: Unfortunately prettier doesn't play well with `templates/docker-compose.yml` with all its blocks.

dessalines (approved):
Thx!

Issue #159, opened 2023-09-02
Add documentation for RHEL installation using the lemmy-almalinux.yml playbook


codyro:
With the merge of #145, the documentation needs to be updated accordingly.

Proposed changes:

* Mention targets the playbook(s) will install on (Debian 10-12, Ubuntu >=22.04 LTS, RHEL 9)
* State which playbook to use for which distribution (EX, `ansible-playbook -i inventory/hosts lemmy-almalinux.yml` for RHEL based distributions)

Issue #171, opened 2023-09-14
Create tags and update documentation 


ticoombs:
Now that we are merging features, we are unfortunately creating more issues because of the breaking changes. 

Moving to tags will allow us to provide a solution so people can update to the latest versions without incurring issues. 

Create tags for `78af02b5e83bc815e549587bb22952256363fc23` (https://github.com/LemmyNet/lemmy-ansible/tree/78af02b5e83bc815e549587bb22952256363fc23) which is the latest "stable" before our variable changes happened (unless I'm wrong)

codyro:
I think what you laid out here, and https://github.com/LemmyNet/lemmy-ansible/issues/169#issuecomment-1718632614 seems reasonable. 

It looks like the maintainers used tags when they updated the `VERSION` file to match the version of Lemmy it represented. We should ensure we don't conflict with their workflows. 

However, I see `@dessalines` thumbed this up already, so we are likely safe to proceed. 

Do you want to do what you laid out in your comment?

I have a semi-related change in https://github.com/LemmyNet/lemmy-ansible/issues/170 we can toss or merge to fix the `vars.yml` issue as well.

ticoombs:
Tag pushed. I probably wont get to doing the documentation today. If you have time that would be welcome. 
I pushed `1.1.0` (without `v`) as I honestly didn't even see the previous tags, so at least this way we can keep a similar pattern while "migrating" everyone to using tags.

codyro:
A new version of Lemmy was released, and a tag was created alongside it: https://github.com/LemmyNet/lemmy-ansible/releases/tag/0.18.5

We need to document our changes soon and probably create a new tag for the latest version of the playbooks/Lemmy. 

I propose we create a new tag (as per semver.org) alongside a `CHANGELOG` file with the changes from `1.1.0` to date. Then I guess we tag alongside the developers when they bump the `VERSION`?

Let me know your thoughts and I can start with that.

ticoombs:
Yes great idea! Please do. 

Also we need to update the initial `git clone` command to checkout the "latest" version.

PR #179, opened 2023-10-04
feat: Version 1.2.0 documentation 

ticoombs:
# Issue

We still tell people to checkout master rather than the latest tag. 

# Solution
Update docs to specify which tag they should  be checking out

# Notes: 
Also added a quick description on what we have changed since 1.1.0, this will suffice until we get that changelog sorted @codyro 

FYI `git checkout $(git describe --tags)` will always get the latest tag. Currently it gets 0.18.5 instead of 1.1.0 as it is technically newer. Which also contains all the "main" updates, such as the `vars` file changes. This may trip people up if we continue to tag with the lemmy version. 
Getting a process that works for the Lemmy devs should be a priority after we push 1.2.0

Closes: #159 #171

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,12 @@ To run this ansible playbook, you need to:
 
 ## Install
 
-1. Clone this repo:
+1. Clone this repo & checkout latest tag
 
    ```
    git clone https://github.com/LemmyNet/lemmy-ansible.git
    cd lemmy-ansible
+   git checkout $(git describe --tags)
    ```
 
 2. Make a directory to hold your config:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # Lemmy-Ansible
 
-This provides an easy way to install [Lemmy](https://github.com/LemmyNet/lemmy) on any server. It automatically sets up an nginx server, letsencrypt certificates, and email.
+This provides an easy way to install [Lemmy](https://github.com/LemmyNet/lemmy) on any server. It automatically sets up an nginx server, letsencrypt certificates, docker containers, and email smtp.
```
codyro, line 3: It's pedantic, but we can mention `pict-rs` too.

codyro (approved):

dessalines (approved):
Thanks! This is going to be really useful.

Issue #180, opened 2023-10-05
Include other lemmy front ends in this ansible.


dessalines:
Now that we have configurable docker-compose.yml (and potentially the same for the nginx), it might be a good idea to add alternate lemmy web front ends, served at their respective names.

The main ones to me seem to be:

- https://github.com/sheodox/alexandrite
- https://github.com/Xyphyn/photon
- https://github.com/aeharding/voyager

cc @erlend-sh

ticoombs:
While it would be simple (I already do this and can upstream my configuration) it adds scope (& complexity). 

If we are wanting to start adding frontends & "extras" re: #174 the ansible code needs to be completely re-written to use proper roles.
ref: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html

nginx, docker, each docker service, letsencrypt, etc would get their own role. Then in the lemmy.yml we call all the roles that need to be installed. It would call the variable file and we would have `role_<rolename>: True` for every role. and if they said true or false that role would be installed, making it a modular installation.

It isn't impossible, it would take some time to re-write which (honestly) will be needed regardless at some point moving forward to better maintain the codebase.

db0:
Might be worth it at this point to make Lemmy an ansible collection and provide the various roles within. It willy massively simplify deployment by users. I can potentially help covert to a collection

dessalines:
@db0 could you link an example or setup of that approach?

db0:
I can provide a PoC sure. It will be a bit hacky but can point out the way to go. As always I have only superficial docker knowledge so I'm not quite sure how I should handle the `docker-compose.yml` (whether it has to be a single file or can be split into multiples is a relevant question for splitting into roles).

db0:
@dessalines PoC provided in #181. It will need further work from you peeps to complete, but this is the general approach. Unfortunately I don't have any more time to work on this right now.

codyro:
That is an excellent example of the direction we should go. @ticoombs suggestion of using variables to determine role inclusion is wise, as it'll allow a single playbook to handle various instances with varying setups.

We need to reach some consensus on the `docker-compose.yml` (EX, https://github.com/LemmyNet/lemmy-docs/issues/274) handling as we have too many varying opinions right now.

erlend-sh:
So just to make sure we’re on the same page, this would make it possible to host Lemmy with an alt-frontend on the instance’s root domain, right?

[...]

PR #181, opened 2023-10-05
Refactor as Ansible Collection and Roles

db0:
As requested here: https://github.com/LemmyNet/lemmy-ansible/issues/180?notification_referrer_id=NT_kwDOAASGQbE3OTQ3OTI1NzQxOjI5NjUxMw#issuecomment-1748721837

I am delivering a proof-of-concept refactor on how you can switch the current repository to ansible collections and roles. For this initial draft, I have simply split the `lemmy` stuff from `pictrs-safety` into individual roles and added the 'photon' role. The expectation is that each component of lemmy can be split into its own role. `backend`, `rust_ui`, `pict_rs`, `pictrs_safety`, `proton`, `alexandrite` etc

The `backend` role contains every lemmy component right now, as I don't have time to do more. But the way I extricated `pictrs_safety` can be used for the others, and to add new frontends follow the approach of photon. I don't know the process to make a frontend the "default", so you'd have to add that yourselves.

I have also made some improvements on your existing ansible code:

- Avoid creating new playbook per distro. One role can handle multiple distros, by using `include_tasks` and `lookup` and only the distinct code per distro needs to be split.
- Switch to using `inventory.yml` instead of `hosts` for better readability 

Note: **This collection is untested**. There's may be some syntax errors etc. But the general conversion is accurate..


codyro:
You stated in the parent issue that you're done working on this at this time. If that's the case, would you mind if I fork this and make some (opinionated) tweaks?

db0:
Feel free to fork it, or work directly on my branch. Yes I'm done working on it for now due to other commitments.

ticoombs:
Closing. Thanks for the PR example!

PR #183, opened 2023-10-06
Config fix for postgres docker defaults

Fmstrat:
Default Docker configurations in Alpine have a different override path, which causes problems if people switch to the Ansible deployment.

Reproduction of the issue:
- Fire up a Lemmy DB in Docker without a volume mount for `/etc/postgresql.conf`
- Use `SHOW config_file` to see the location of the config file (it will be in `/var/lib/data`)
- Take down the DB, and add a volume mount for `/etc/postgresql.conf`
- Use `SHOW config_file` to see the location of the config file (it will still be in `/var/lib/data`)

Also, official docker instructions for Postgres is to use `/etc/postgresql/postgresql.conf`, though this is irrelevant given this PR.

To force the use of `/etc/postgresql.conf` for if users switch from bare mode to Ansible, the compose file was edited to include a forced `command`.

In addition, the default for PG15 alpine `listen_address` is `localhost`, not `*`, which is required for the `lemmy` container to connect. This default is overwritten by the docker container normally, but is not included in the tuning samples provided.

***NOTE***: It is highly possible there are a number of smaller self-hosted instances who have no idea their poor performance is because PG is not properly tuned, as the defaults are pretty meager.

### DIFF
--- a/templates/docker-compose.yml
+++ b/templates/docker-compose.yml
@@ -101,6 +101,7 @@ services:
       - ./volumes/postgres:/var/lib/postgresql/data:Z
       - ./customPostgresql.conf:/etc/postgresql.conf
     restart: always
+    command: postgres -c config_file=/etc/postgresql.conf
     logging: *default-logging
 
   postfix:
--- a/examples/customPostgresql.conf
+++ b/examples/customPostgresql.conf
@@ -28,3 +28,6 @@ temp_file_size=1GB
 synchronous_commit=off
 # This one shouldn't be on regularly, because DB migrations often take a long time
 # statement_timeout = 10000
+
+# Listen beyond localhost
+listen_addresses = '*'


templates/docker-compose.yml
```diff
@@ -101,6 +101,7 @@ services:
       - ./volumes/postgres:/var/lib/postgresql/data:Z
       - ./customPostgresql.conf:/etc/postgresql.conf
     restart: always
+    command: postgres -c config_file=/etc/postgresql.conf
```
codyro, line 104: Any reason we shouldn't just mount the `postgresql.conf` to the expected default `/var/lib/postgresql/data/postgresql.conf` instead of forcing our own?

If possible, I'd prefer to have only one copy of `postgresql.conf` on the VM to help prevent future confusion.

Fmstrat: It's not detailed in the official docs, but I don't see why we couldn't. Let me do some testing to confirm a fresh DB start wouldn't try to overwrite it, and I'll update the PR accordingly.

Moving this weekend, so might not be until Monday.

Fmstrat: And now we know why you don''t mount it there:

```
initdb: error: directory "/var/lib/postgresql/data" exists but is not empty
initdb: hint: If you want to create a new database system, either remove or empty the directory "/var/lib/postgresql/data" or run initdb with an argument other than "/var/lib/postgresql/data".
```

Postgres won't create the proper file system in `/var/lib/postgresql/data` because a file exists in the directory already.

One option would be to use the recommended `/etc/postgresql/postgresql.conf`, but honestly, even with this way it leaves the door open to an administrator using a default config to start and not realizing it won't override at a future time, and in this particular situation it would still result in two files. I would suggest merging as-is to ensure the "use of wrong config" part of this doesn't happen, but let me know if you'd like me to make a change.

codyro: What a funny chicken-and-egg situation. Thanks for checking. I concur. The path is moot at that point, so we may as well keep it as-is.

codyro:
Taking a quick look, I think we may run into issues with the `/etc/postgresql.conf` being mounted with permissions the `postgres` user cannot read:

```
/ # ls -lh /etc/postgresql.conf
-rw-------    1 1000     1000         733 Oct  2 18:12 /etc/postgresql.conf

/ # grep postgres /etc/passwd
postgres:x:70:70:Linux User,,,:/var/lib/postgresql:/bin/sh
```

```
2023-10-07 20:01:21.001 GMT [1] LOG:  could not open configuration file "/etc/postgresql.conf": Permission denied
2023-10-07 20:01:21.001 GMT [1] FATAL:  configuration file "/etc/postgresql.conf" contains errors
```

The playbooks are distributing the postgres configuration with `1000:1000` ownership and `0600` permissions (https://github.com/LemmyNet/lemmy-ansible/blob/main/lemmy.yml#L222-L224). I see no reason why this is needed, so we could either change the ownership to `70:70` and keep the `0600` permissions or just set the `user` and `owner` to `root` with `0644` permissions (IMO the most logical). 

Could you fix this in your PR?

Fmstrat:
Happy to do it, while I don't deploy lemmy this way (I just happened to notice it) I'm prretty familiar with ansible. I'll knock it out after the move. And thanks!

ticoombs:
Thanks @Fmstrat ! I've been meaning to fix this up. As i've helped solve this on the matrix rooms multiple times.

ticoombs (approved):

dessalines (approved):

PR #185, opened 2023-10-09
fix: force no become on localhost (fixes: #178)

ticoombs:
# Issue 

When using `--become` this tells ansible to always get root (or whatever become_user is). This includes when we delegate
Initial https://github.com/LemmyNet/lemmy-ansible/pull/178 was insufficient

# Solution

when we delegate make sure we `become: false`

# Reference

#ansible-chat in matrix room (https://matrix.to/#/!XunuLxRTzInMzxQxdS:discuss.online/$f9EsmBIH1UL9O9kmuUhhprsjfhPH73xd7ac1PU_JlYc?via=discuss.online&via=matrix.org&via=nope.chat)

improvement on #178

Issue #177, opened 2023-10-01
ERROR! 'notify' is not a valid attribute for a Block


pixlguru:
Update failed 0.18.4 -> 9.18.5

I followed the update instructions, 
- renamed postgres.psk
- copied vars.yml

`
ansible-playbook -i inventory/hosts lemmy.yml --become`

```
ERROR! 'notify' is not a valid attribute for a Block

The error appears to be in '/home/user/workspace/folder/lemmy-ansible/lemmy.yml': line 190, column 11, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


        - name: Set up nginx sites-enabled symlink
          ^ here

```

codyro:
What version of Ansible are you on (`ansible --version`)? Using `notify` with a `block` is supported on the latest versions:

https://docs.ansible.com/ansible/latest/reference_appendices/playbooks_keywords.html#block

Can you update Ansible and see if it works for you?

pixlguru:
> Can you update Ansible and see if it works for you?

It does, thanks a lot!

omalaspinas:
I have the same error when upgrading to lemmy-ansible version 1.2.0. Working on Ubuntu 22.04. Ansible 2.10.8.

Last time I made an update was a bit after the release of the lemmy version to 0.18.4 (don't know the exact commit though).

Any idea?

codyro:
It looks like support for this was added after Ansible 2.10. If you upgrade to the latest version of Ansible, this will work. I'll have the documentation updated with the appropriate Ansible version once I confirm when this was supported.

You can install the latest version quickly from pip by running `python3 -m pip install ansible` (this should grab `2.15`, which does support `notify` usage in `block`s.

codyro:
Looks like it was added in 2.11 (Ansible 4).

Ref: 
* https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-community-changelogs
* https://github.com/ansible-community/ansible-build-data/blob/main/4/CHANGELOG-v4.rst#L7552

omalaspinas:
You are right. It's working. Thank  you!

PR #188, opened 2023-10-12
Update ansible requirements, remove `become` for localhost delegated tasks

codyro:
Ansible did not support `notify` in `block`s until version `2.11.0`. To avoid having users receive an unfriendly error (https://github.com/LemmyNet/lemmy-ansible/issues/177), let them know that the version of Ansible they're running is too old. 

# Changes

* Update `README.md` to add `ansible-core >= 2.11.0` requirement
* Add assertions to `lemmy.yml` & `lemmy-almalinux.yml` playbooks to verify the Ansible version before continuing to run. This will help in instances such as the one in #177 
* Add `become: false` to tasks running on `localhost` to the `lemmy-almalinux.yml` (brings in-line with `lemmy.yml`. Refs #185)


dessalines (approved):
Thx!

ticoombs (approved):
lgtm

Issue #191, opened 2023-10-15
issue with postgree password with playbook 1.2.0 ?


tifriis208:
Hello,

I try a from scratch install of lemmy on ubuntu 22.04.
Playbook runs withtout issue, but when I try to connect to my instance, I get the following message on my browser:
_Erreur !
Une erreur s'est produite sur le serveur. Essayez d'actualiser votre navigateur. Si cela ne fonctionne pas, revenez plus tard. Si le problème persiste, vous pouvez demander de l'aide à la [Communauté d'assistance Lemmy](https://lemmy.ml/c/lemmy_support) ou à la [Groupe Matrix Lemmy](https://lemmy.ml/c/lemmy_support)._

In my logs, I got :
```
root@xxx:/srv/lemmy/xxxxxx# docker-compose logs -f lemmy
Attaching to lienslowtechfr_lemmy_1
lemmy_1     | thread 'main' panicked at 'Error connecting to postgres://lemmy:XBBmfAp9PNLJuxv56eWY@postgres:5432/lemmy: could not connect to server: Connection refused
lemmy_1     | 	Is the server running on host "postgres" (172.18.0.2) and accepting
lemmy_1     | 	TCP/IP connections on port 5432?
lemmy_1     | ', crates/db_schema/src/utils.rs:217:56
lemmy_1     | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
lemmy_1     | thread 'main' panicked at 'Error connecting to postgres://lemmy:XBBmfAp9PNLJuxv56eWY@postgres:5432/lemmy: could not connect to server: Connection refused
```

Seems like an issue to connect to postgree, maybe a password issue ?

Never have this issue before.

Thanks :)

codyro:
It doesn't look like the PostgreSQL VM is running. Can you run the following and see if it's running and or constantly starting/crashing?

`docker ps -a | grep postgres`

You can also check the Docker logs for the instance by pulling the VM ID from the above command and running:

`docker logs <VMID>`

codyro:
Actually,

Can you clarify if you checked out the `main` branch or the `1.2.0` tag? This should already be fixed and or fixed in `1.2.1` (https://github.com/LemmyNet/lemmy-ansible/pull/183) when it gets pushed.

Please check this issue and see if it helps: https://github.com/LemmyNet/lemmy-ansible/issues/190

mszaro:
I'm also seeing the same issue, with `main` checked out:

```
[...]

PR #192, opened 2023-10-17
Remove invalid `temp_file_size` line in `customPostgresql.conf`

codyro:
Now that customPostgresql.conf is being correctly used, it looks like it has an invalid value which causes PostgreSQL not to start ([#191](https://github.com/LemmyNet/lemmy-ansible/issues/191)). This PR removes the offending line which allows PostgreSQL to start.

```
2023-10-17 01:32:35.372 GMT [1] LOG:  unrecognized configuration parameter "temp_file_size" in file "/etc/postgresql.conf" line 27
2023-10-17 01:32:35.372 GMT [1] FATAL:  configuration file "/etc/postgresql.conf" contains errors
```

@dessalines @Nutomic  If you know what you were using this line for and or what it can be replaced with, let me know.

### DIFF
--- a/examples/customPostgresql.conf
+++ b/examples/customPostgresql.conf
@@ -24,7 +24,6 @@ max_parallel_workers = 4
 max_parallel_maintenance_workers = 2
 
 # Other custom params
-temp_file_size=1GB
 synchronous_commit=off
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Hrm, they must have removed that one. NBD.

PR #194, opened 2023-10-20
Redirecting version to nodeinfo.

dessalines:
Apologies, we decided to go another way with it and just redirect it in nginx.


Nutomic:
We could still do the redirect in Rust and keep the nginx block like your previous PR.

Issue #84, opened 2023-06-13
Wrong docker image pulled on arm64 systems.


justdan6:
```lemmy_docker_image``` and ```lemmy_docker_ui_image``` vars in ```lemmy.yml``` should change to the arm64 image if the server is an arm64 based system.

Currently it's defaulted to amd64 which causes ```exec format error``` in the ```lemmy``` and ```lemmy-ui``` images.

**Comment by justdan6** [comment]
My current fix is to just hard code the version - 
```
lemmy_docker_image: "dessalines/lemmy:0.17-linux-arm64"
lemmy_docker_ui_image: "dessalines/lemmy-ui:0.17-linux-arm64"
```

tifriis208:
A big thanks for sharing this. I got the same issue and it solved it :)

Do you know how we can update to lemmy 0.17.4 on arm64 ?
I also noticed that picture are no more display, do you have this issue too ?

For me, this is issue is related to this issue https://github.com/LemmyNet/lemmy-ansible/issues/86 and quite related also to this issue https://github.com/LemmyNet/lemmy-ansible/issues/67

justdan6:
It looks like you'd need to build your own image to go higher than 17.3 if you want to use docker. The most recent arm64 build is 17.3 - https://hub.docker.com/r/dessalines/lemmy/tags?page=1&name=arm64

ravermeister:
> It looks like you'd need to build your own image to go higher than 17.3 if you want to use docker. The most recent arm64 build is 17.3 - https://hub.docker.com/r/dessalines/lemmy/tags?page=1&name=arm64

Is anyone currently building one?

paulo-roger:
> > It looks like you'd need to build your own image to go higher than 17.3 if you want to use docker. The most recent arm64 build is 17.3 - https://hub.docker.com/r/dessalines/lemmy/tags?page=1&name=arm64
> 
> Is anyone currently building one?

Apparently we are on our own with this.

I am trying to learn how to do it myself now.

omalaspinas:
I managed to build an arm docker image for lemmy. I have problems with lemmy-ui though. Here is my Dockerfile for lemmy-18:

```docker
FROM blackdex/rust-musl:aarch64-musl as builder
WORKDIR /app
ARG CARGO_BUILD_TARGET=aarch64-unknown-linux-musl

# This can be set to release using --build-arg
ARG RUST_RELEASE_MODE="debug"

COPY . .

# Build the project
    
[...]

PR #197, opened 2023-10-25
#84 : update documentation with architecture requirement

tifriis208:


codyro:
Thank you! Would you mind running prettier over this so it passes the CI checks? 

https://woodpecker.join-lemmy.org/repos/130/pipeline/212/3

dessalines (approved):

ismailkarsli:
Everything works on ARM with @ubergeek77's images. May I ask why Lemmy's own images don't working on ARM?

https://github.com/ubergeek77/lemmy-docker-multiarch

codyro:
> Everything works on ARM with @ubergeek77's images. May I ask why Lemmy's own images don't working on ARM?

It's just a time and resource issue. Check out the following issues for more context (and feel free to contribute!):

https://github.com/LemmyNet/lemmy/issues/3102
https://github.com/LemmyNet/lemmy/pull/4004

ismailkarsli:
> > Everything works on ARM with @ubergeek77's images. May I ask why Lemmy's own images don't working on ARM?
> 
> It's just a time and resource issue. Check out the following issues for more context (and feel free to contribute!):
> 
> [LemmyNet/lemmy#3102](https://github.com/LemmyNet/lemmy/issues/3102) [LemmyNet/lemmy#4004](https://github.com/LemmyNet/lemmy/pull/4004)

Believe me, If I could, I would :) Configuring Rust to compile is just hard to me.

I guess the difference here is Lemmy team wants final OS to be Alpine and ubergeek77 is using Debian. I got my answer, thanks.

ubergeek77:
Just wanted to chime in since I was tagged - I originally tracked whatever the official Lemmy images were using in the past. I haven't touched the backend image in 4 months, and at that time they were using Debian. I'll probably go through and update it to use Alpine, and have parity with the official images. I'm not using Debian for any specific reason.

There is really nothing special about my Dockerfiles, I simply remove the cross-compilation aspects and add fixes to ensure the images will build on any architecture. My repository is configured to launch cheap AWS VMs to do this bare metal building in lieu of cross compilation. Real ARM VMs are used for the two ARM targets.

This has been discussed before, and the answer was that the existing Lemmy Woodpecker CI/CD, which runs 24/7 instead of on-demand, does not support ARM. It is not known why remaining on Woodpecker is a hard requirement given this limitation, especially if an on-demand CI/CD stack on AWS would be very inexpensive to run and well within the Lemmy project's publicly listed funding. It could also be used to augment the existing Woodpecker infrastructure and only used for stable builds, and doing so would cost the Lemmy project a maximum of ~$2 per month with little to no management overhead (I haven't even ***looked*** at my AWS console in 3 months, and my images continue to build automatically).

This is not really something anyone can "contribute," as it goes beyond code in any specific repo, and would have to be an operational choice the Lemmy project makes. I won't be opening any PRs, but my approach is fully open source and the maintainers are more than welcome to adopt it if they want.

codyro:
> It is not known why remaining on Woodpecker is a hard requirement 

I cannot speak for the maintainers. However, I assume they're trying to keep the CI/CD pipeline completely FOSS and not rely on a proprietary system/GH. This makes the pipeline agnostic and easy to move elsewhere if they opt to stop using GitHub/if they would prefer to move to something on-premise, etc.

I have an idea I would like to test to see if I can get this to build cleanly as-is; however, if it's too much of a pain, I'll sponsor a 24/7 ARM runner for the pipeline to make things easier (not a RPI ;)), as it's something people want. @dessalines / @Nutomic, if you would like to ping me somewhere, we can coordinate getting that going by tomorrow (assuming my idea won't work, which I'll verify later tonight or in the AM). It doesn't resolve the "building from ARM" issue, but it is a step in the right direction.

For reference, I love how you approached things in your repository and will be re-using some of the logic for some of my projects, so thank you for that :)

Nutomic:
There is a new PR to get x86 to ARM cross compilation working: https://github.com/LemmyNet/lemmy/pull/4142. Once that is merged the issue should be resolved. Having all builders on x86 should be easier to manage. If that doesnt work we can also go the route via ARM builder, but it looks like that would also require fixes to the dockerfile and Id rather not have two conflicting PRs for this.

ticoombs:
@Nutomic Would you confirm for me now that https://github.com/LemmyNet/lemmy/pull/4142 & https://github.com/LemmyNet/lemmy/pull/4202 have been merged that we are now ARM64 & AMD64 compatible?

Nutomic:
Yes that is correct, although Im not sure if ARM builds have been tested in production yet. But it would fit better in the requirements list instead of this table.

PR #202, opened 2023-12-03
Fix installation playbooks running on incompatible OSs

sanatsathaye:
The Almalinux playbook runs on Ubuntu and Debian, because the current `fail` logic is only succesful when _both_ these conditions are true:

1. Not RHEL
2. Version < 9

So, it will only fail on some obscure system like Ubuntu 8 (side note, if you _are_ trying to install Lemmy on a server with Ubuntu 8 I need answers).

To fix this I changed the `fail` to an `assert` and inverted the conditions. Changing the current "and" clause in the `fail` to an "or" would also make this work but IMO an assertion works better here.

### DIFF
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -65,11 +65,11 @@
     lemmy_port: "{{ 32767 | random(start=1024) }}"
   tasks:
     - name: Ensure target system is >= EL9
-      ansible.builtin.fail:
-        msg: "This playbook requires Enterprise Linux 9 or greater"
-      when:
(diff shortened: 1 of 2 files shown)


codyro:
What a fun catch! When I created the playbook, I tried to avoid it running on unsupported RHEL systems. I hadn't considered anyone would try the RHEL/Alma playbook on Ubuntu/Debian systems. 

I'll get this checked and merged once I'm home :).

codyro (approved):

codyro:
Thanks!

(Thanks @ticoombs for tagging this, I promise I'm trying to be mindful of it!)

PR #204, opened 2023-12-15
Adding a 1GB shm size for postgres, to prevent shared space error.

dessalines:
Context: https://stackoverflow.com/a/56754077/1655478


ticoombs (approved):
Can confirm we need this.

ticoombs:
:rocket: Merging

PR #210, opened 2023-12-18
feat: postgres backup to filesystem

ticoombs:
# Problem

Documented backup solutions [(here)](https://join-lemmy.org/docs/administration/backup_and_restore.html) are inefficient and cause all data from the backup to be piped via the docker console/output rather than directly to a file.

Outputting directly to a file results in atleast a 4x speedup of backups, and does not require docker console to proxy the whole database.

(to be merged after #209)

### DIFF
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -150,6 +150,8 @@
           owner: "root"
         - path: "{{ lemmy_base_dir }}/{{ domain }}/volumes/"
           owner: "root"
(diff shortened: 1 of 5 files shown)


Nutomic:
I think this would fit better in the documentation. Then it will also benefit people who dont use Ansible.

https://github.com/LemmyNet/lemmy-docs/blob/main/src/administration/backup_and_restore.md

dessalines:
Agree, we could just link to that backup and restore page on this readme, if you think that'd be useful.

Issue #229, opened 2024-02-28
Alma linux 9 DO droplet doesn't run through


RussellTaylor83:
Hello,

I thought I would stick with a RHEL distro as I use it for work. I have used a Digital Ocean Alma Linux 9 droplet, and I can ssh to it as root. There seem to be a few hurdles, and I then have postgres file permissions failures I can't overcome. I'm not sure if it's a problem with what I'm doing or something to try and resolve. I run the lemmy-almalinux.yml and get 

> RUNNING HANDLER [Reload nginx] ****************************************************
fatal: [root@mydomain]: FAILED! => {"changed": false, "msg": "Unable to start service nginx: Job for nginx.service failed because the control process exited with error code.\nSee \"systemctl status nginx.service\" and \"journalctl -xeu nginx.service\" for details.\n"}

This is because the port is already in use by another nginx process, that must have been started as part of the ansible script.

I pkill -f nginx and run again... next time it runs straight through. I get a 502, so checking podman-compose logs it shows the postgres container lasting a few seconds with many logs saying

> 26d97800a297 PostgreSQL Database directory appears to contain a database; Skipping initialization
26d97800a297
26d97800a297 postgres: could not access the server configuration file "/etc/postgresql.conf": Permission denied

I have attempted chmod'ing the customPostgresql.conf file, moving it to /tmp and mounting from there, and can't seem to get past this permissions error.

If I docker compose down and then up, the postgres section of the logs look like this (I have replaced the domain I'm using with "mydomain")

[...]

PR #231, opened 2024-03-09
AlmaLinux/RHEL 9 fixes

codyro:
Ensures `lemmy-almalinux.yml` runs on a clean AlmaLinux/RHEL9 instance.

- Move internal nginx resolver to variable/conditional based off `os_family`
- Modify SELinux in the `lemmy-almalinux.yml` playbook to ensure clean run
- Modify `docker-compose.yml` template to ensure SELinux label gets applied (`:Z`)
- ~~Start & Enable nginx earlier in the playbook to avoid conflict~~ (Already merged in https://github.com/LemmyNet/lemmy-ansible/pull/230)
- Refs https://github.com/LemmyNet/lemmy-ansible/issues/229

### DIFF
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -239,8 +239,6 @@
       tags:
         - nginx
 
-
-
     # TODO: Check if this is necessary with EL & podman
     # - name: Copy docker config
     #   ansible.builtin.copy:
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -129,6 +129,17 @@
       tags:
         - firewalld
 
+    - name: Start and enable nginx.service
+      ansible.builtin.systemd:
+        name: nginx.service
+        state: started
+        enabled: true
+      tags:
+        - certbot
+        - certbot_initial
+        - nginx
+        - ssl
+
     # TODO: certbot logic needs to be re-worked
     - name: Request initial letsencrypt certificate
       ansible.builtin.command: certbot certonly --nginx --agree-tos --cert-name '{{ domain }}' -d '{{ domain }}' -m '{{ letsencrypt_contact_email }}'
@@ -227,6 +238,8 @@
       notify: Reload nginx
       tags:
         - nginx
+
+
 
     # TODO: Check if this is necessary with EL & podman
     # - name: Copy docker config
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -13,12 +13,6 @@
         fail_msg: "This playbook requires Ansible 2.11.0 or higher"
       become: false
 
-    - name: Disable SELinux
-      ansible.posix.selinux:
-        state: disabled
-      tags:
-        - selinux
-
     # This is not needed for this playbook as it predates its existence
     # But we're keeping it for funsies :)
     - name: Check lemmy_base_dir
@@ -134,6 +128,15 @@
       when: "'firewalld.service' in ansible_facts.services and ansible_facts.services['firewalld.service'].state == 'running'"
       tags:
         - firewalld
+
+    - name: Adjust SELinux to allow HTTPD scripts and modules to connect to the network
+      ansible.posix.seboolean:
+        name: httpd_can_network_connect
+        state: true
+        persistent: true
+      tags:
+        - nginx
+        - selinux
 
     - name: Start and enable nginx.service
       ansible.builtin.systemd:
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -13,6 +13,12 @@
         fail_msg: "This playbook requires Ansible 2.11.0 or higher"
       become: false
 
+    - name: Disable SELinux
+      ansible.posix.selinux:
+        state: disabled
+      tags:
+        - selinux
+
     # This is not needed for this playbook as it predates its existence
     # But we're keeping it for funsies :)
     - name: Check lemmy_base_dir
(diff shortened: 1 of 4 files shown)


lemmy-almalinux.yml
```diff
@@ -13,6 +13,12 @@
         fail_msg: "This playbook requires Ansible 2.11.0 or higher"
       become: false
 
+    - name: Disable SELinux
+      ansible.posix.selinux:
+        state: disabled
+      tags:
+        - selinux
+
```
ticoombs, line 21: I'm not sure disabling selinux is the correct (forced?) solution for this. But again, I do not have enough redhat experience for this. So if you think this is the best solution then i'll roll with it.

codyro: I kept waffling on this, too. It is less than ideal, and I wouldn't say I like it, but to get the playbooks to run cleanly, SELinux needs to be disabled/set to permissive (or modified outside the scope of this playbook). I decided to disable it for the same reason I thought it was okay to punch 80/443 into the firewalls--people expect to run these playbooks cleanly out-of-the-box (see the referenced issue, for example). 

My initial approach was to detect if SELinux is enabled, and if so, prompt the user, warning them it might fail until they disable it. Alternatively, a prominent warning/addition to the README may be sufficient (or a combination). 

Do you have any thoughts on this more passive approach? The user experience is worse, but it’s more “proper” (and secure) to leave SELinux alone and have the user manually turn it off.

Nutomic: If you disable selinux then I would at least expect that you disable it at the end of the playbook. And then its still possible that an error is thrown somewhere and it stays disabled which doesnt sound good.

codyro: > If you disable selinux then I would at least expect that you disable it at the end of the playbook

The playbook & Lemmy deployment will not work with SELinux enabled/enforcing. It must be disabled at the start to ensure clean deployment & functioning. 

> And then its still possible that an error is thrown somewhere and it stays disabled which doesnt sound good.

I think I missed an important reason I took this approach—Lemmy being deployed via these playbooks *will not* function without additional tweaking of SELinux policies and contexts, which I believe is outside the scope of this playbook. 

While it’s not the best, this is why some cloud providers ship RHEL9 images with it disabled by default, even if differs from upstream—it causes issues for laypeople who don’t understand what’s going on as the system behaves bizarrely (in their eyes). 

If we want to be cautious, I think the best route is to state that SELinux needs to be disabled/permissive to run Lemmy via these playbooks. 

I’ll look today to see how much tweaking is needed to get these playbooks & Lemmy running with SELinux enabled/enforcing. Still, I think we’ll see more issues in the long run with all of the variables SELinux throws in that we would need to account for. If we want to go this route, I don’t mind supporting it. It will likely require manual work from the server operator/proper documentation (which might be beneficial to toss in the official Lemmy docs for RHEL). 

I’ve reiterated this point—these playbooks should be as frictionless as possible. In its current state, it is not possible to get Lemmy going/installed without modifications to SELinux, which will open a whole new can of worms.

dessalines: I agree with keeping them as frictionless as possible.

dessalines (approved):

lemmy-almalinux.yml
Nutomic: Alright if its that much effort then lets keep it simple.

codyro:
Thanks for keeping me on my A-game @ticoombs & @Nutomic! I spent some time to see if this could be handled better, and we can!

When the `docker-compose.yml` file was made, it didn't get configured consistently and was missing an option on a couple of container mounts that configure the SELinux label to ensure they work as expected. (https://github.com/LemmyNet/lemmy-ansible/pull/231/files#diff-d3d33979648a9836685e314d1864481561d5130a294cce04be84b03c29abfb08)

The other issue was that `nginx` was getting denied, making it unable to proxy to the containers. (https://github.com/LemmyNet/lemmy-ansible/pull/231/files#diff-67f417e813ca24645bdcb19c0b344055cc0b48d114204e7026827b634cdc8815R132-R139)

@ticoombs Before we merge (or at least tag a future release), would you mind pinging me or assigning me to the PR so I can ensure the RHEL side of things looks okay? The most minor things can cause it to break 😓 .

RussellTaylor83:
Hello, thanks for working on this.

I gave this a quick run on Digital Ocean / Alma 9 and got:

> TASK [Distribute docker/podman templates] *****************************************************************************************************************************************
changed: [root@123.123.123.123] => (item={'src': 'templates/docker-compose.yml', 'dest': '/srv/lemmy/mydomain.uk/docker-compose.yml', 'mode': '0600'})
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleUndefinedVariable: 'nginx_internal_resolver' is undefined. 'nginx_internal_resolver' is undefined
failed: [root@123.123.123.123] (item={'src': 'templates/nginx_internal.conf', 'dest': '/srv/lemmy/aidriven.uk/nginx_internal.conf', 'mode': '0644'}) => {"ansible_loop_var": "item", "changed": false, "item": {"dest": "/srv/lemmy/aidriven.uk/nginx_internal.conf", "mode": "0644", "src": "templates/nginx_internal.conf"}, "msg": "AnsibleUndefinedVariable: 'nginx_internal_resolver' is undefined. 'nginx_internal_resolver' is undefined"}

I used the same config as I have for my Debian deployment that works. I haven't time at the moment to dig into this but may be able to later.

Thanks

codyro:
I added a new variable in that PR that you'll want to add to your `vars.yml` for your install, which will fix this error:

https://github.com/LemmyNet/lemmy-ansible/pull/231/files#diff-fadf44a49d340433e809fac78150211bc15b8047e299b14bf891197a2f8a54fdR59

RussellTaylor83:
I should have read the PR, my bad. 

That ran through fine, and from a basic test seems to work ok, thank you.

Issue #189, opened 2023-10-15
Exceeding letsencrypt cert limit


akzmonster:
I have been trying to get this thing working for months now and I've had to troubleshoot so many things that are missing from this guide. In doing so, I've had to reset my local machine multiple times and wipe out my VPS every time. But I suppose this has requested a brand new letsencrypt cert every time, and I just hit a limit. I've been locked out for 168 hrs from requesting another.

Do you have any way of implementing a letsencrypt cert retention mechanism so that I can troubleshoot without these limitations?

How would I be able to use an existing cert in a playbook?

Nothing4You:
this is currently not configurable in the playbook, but you could comment out these two blocks:

https://github.com/LemmyNet/lemmy-ansible/blob/1.2.0/lemmy.yml#L136-L139

https://github.com/LemmyNet/lemmy-ansible/blob/1.2.0/lemmy.yml#L246-L251

and store the cert in `/etc/letsencrypt/live/$domain/fullchain.pem` along with the private key in `/etc/letsencrypt/live/$domain/privkey.pem`.

if you want to deploy the cert through ansible, you could add `fullchain.pem` and `privkey.pem` to the `files` folder and add the following snippet to `lemmy.yml` after the `Request initial letsencrypt certificate` task:

```yaml
    - name: Create folder for certificates
      ansible.builtin.file:
        path: "/etc/letsencrypt/live/{{ domain }}"
        state: directory
        owner: root
        group: root
        mode: "0755"

    - name: Deploy certificate files
      ansible.builtin.copy:
        src: "files/{{ item }}"
        dest: "/etc/letsencrypt/live/{{ domain }}/{{ item }}"
        owner: root
        group: root
        mode: "0644"
      loop:
        - fullchain.pem
        - privkey.pem
```

I'm not sure about the production file permissions in this case, but for testing this shouldn't matter.

codyro:
We should offer the option to use your own SSL certificate and a variable to enable/disable certbot from running. 

If we're feeling fancy, a variable to have certbot run but using `--dry-run` could be helpful for instances like these as well.

ticoombs:
@akzmonster 

I know some time has passed, but would you mind explaining what is missing in our documentation that caused the resulting issue in your case? 
The letsencrypt certificate limit is symptom, not the cause. If we instead try to solve our documentation issues then that would be a better way to ensure no-one in the future has the same issues.

ticoombs:
[...]

PR #236, opened 2024-04-06
Add missing ansible tags

sanatsathaye:
Added tags to `lemmy.yml` and `lemmy-almalinux.yml`. Hope they look fine, pls suggest if there's any that can be changed. 

Also, this will kinda sorta help with #189 because you can run something like `ansible-playbook lemmy.yml --skip-tags ssl` that will effectively bypass everything related to ssl in case you want to reuse existing certificates (or your setup is like mine where I test out the files within WSL/vagrant without a valid domain).

Honestly speaking I wouldn't recommend running the playbook solely relying on tags at all unless you're extremely familiar with both the playbook and ansible, but it can be used for cases like I mentioned.

I don't think `uninstall.yml` needs tags because it completely works on user input so I didn't bother.

### DIFF
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -172,6 +172,8 @@
           owner: "root"
         - path: "{{ lemmy_base_dir }}/{{ domain }}/volumes/pictrs/"
           owner: "991" # Matches docker-compose UID in docker-compose.yml
+      tags:
+        - directories
 
     - name: Set lemmy_port fact
       ansible.builtin.set_fact:
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -15,11 +15,15 @@
           - "ansible_version.full is version('2.11.0', '>=')"
         fail_msg: "This playbook requires Ansible 2.11.0 or higher"
       become: false
+      tags:
+        - always
 
     - name: Check lemmy_base_dir
       ansible.builtin.fail:
         msg: "`lemmy_base_dir` is unset. if you are upgrading from an older version, add `lemmy_base_dir=/lemmy` to your inventory file."
       when: lemmy_base_dir is not defined
+      tags:
+        - always
 
     - name: Check for legacy passwords/postgres file
       delegate_to: localhost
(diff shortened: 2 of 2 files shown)


sanatsathaye:
@codyro did most of the work with tags already in the Alma playbook, so I've re-used as much as I can 😄

ticoombs:
Is there a reason we've gone so descriptive with these? ie; can you let us know how we'd want to use them?

- `--skip-tags directories ` (to speed up?)
- `--tags always --tags docker` (to install docker?)
- `--skip-tags dependencies`  (speed up?)

If we have these tags we should also have some documentation on how to _use_ them efficiently. Otherwise I can see people running into problems.

I understand ssl/certbot/initial as users could skip that and that could help solve #189 but i'm not sure of the others.

(We'll most likely accept it as it is, but if we can have some guidance on how we could utilise these tags that would be great)

codyro:
> Is there a reason we've gone so descriptive with these? ie; can you let us know how we'd want to use them?
> 
> 
> 
> - `--skip-tags directories ` (to speed up?)
> 
> - `--tags always --tags docker` (to install docker?)
> 
> - `--skip-tags dependencies`  (speed up?)
> 
> 
> 
> If we have these tags we should also have some documentation on how to _use_ them efficiently. Otherwise I can see people running into problems.
> 
> 
> 
> I understand ssl/certbot/initial as users could skip that and that could help solve #189 but i'm not sure of the others.
> 
> 
> 
> (We'll most likely accept it as it is, but if we can have some guidance on how we could utilise these tags that would be great) 
> 
> 

Regarding the playbook I made, I tend to do it for development reasons (which sometimes can be helpful in production). I don't think it's harmful to have, though.

I used the `always` tag because it's a special use one: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tags.html#special-tags-always-and-never

It lets you ensure some tasks always run even when limiting them to a subset via `--tags`, providing mission critical tasks run regardless of what's passed. I assume that’s what @ was using as an example/baseline.

dessalines:
I'll defer to @codyro 's recommendations on this one, but simpler is better.

ticoombs (approved):

Issue #227, opened 2024-02-05
On next release, include upgrade instructions for postgres 15 -> 16


dessalines:
This is unfortunately necessary due to a postgres 15 memory leak: https://github.com/LemmyNet/lemmy/issues/4406

This PR has the file: https://github.com/LemmyNet/lemmy/pull/4426/files#diff-4bb683dcf59cc3a55cf3b7c6f0b6a202149806f28f6ab704a83050d7d3aabf6c

Once that's merged, we need to make sure to  have instructions for the upgrade on the next lemmy-ansible release.

ticoombs:
So this also removes the need to have `jit = no` in our customPostgres file too.

dessalines:
I don't know enough about that setting to say if we should keep / remove, but I'd say pry remove.

ticoombs:
lemmy.cafe dev/admins "confirmed" that the leak no longer is present when using `jit=no` so we can get rid of that too.

Issue #211, opened 2023-12-18
Recommended/default config should include pg_stat_statements and auto_explain


phiresky:
It's annoying to get vague reports about db perf issues without basic tools to get more info. I thought these were already part of the recommended installation but doesn't look like it

They don't affect performance (afaik) so shoudl be fine to always have active

ticoombs:
I gather you mean these 3 configuration options? https://github.com/LemmyNet/lemmy/issues/4282#issuecomment-1860521011

From the [auto_explain documentation](https://www.postgresql.org/docs/current/auto-explain.html);  (Emphasis is mine)
> More typical usage is to preload it into some or all sessions by including auto_explain in [session_preload_libraries](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-SESSION-PRELOAD-LIBRARIES) or [shared_preload_libraries](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-SHARED-PRELOAD-LIBRARIES) in postgresql.conf. Then you can track unexpectedly slow queries no matter when they happen. **Of course there is a price in overhead for that.**

----

I agree it would be a good idea, especially in helping to diagnose these issues everyone is happen. . What about if we have a commented debugging section in `customPostgres.conf` with sane-defaults so that admins can add / remove at will. Do you think that would be an acceptable middleground?

phiresky:
According to the same docs, most of the overhead only happens if log_analyze is set to true:

> Turning off auto_explain.log_timing ameliorates the performance cost, at the price of obtaining less information

It's hard to find a lot of specific information, but according to https://www.pgmustard.com/blog/auto-explain-overhead-with-timing the overhead may be in the order of 1% (with default settings). That sounds to me like a good cost for getting useful info out of instance admins

Issue #195, opened 2023-10-23
Consider enabling nginx caching


Nutomic:
Lemmy now sends proper cache-control headers so that responses for unauthenticated users can be cached. This would help to reduce resource usage on the server. However it could also lead to weird issues like stale responses. I previously attempted to add caching in https://github.com/LemmyNet/lemmy-ansible/pull/75, but back then it was quite hacky as Lemmy didnt set cache-control. 

You can enable nginx caching by adding the following at the top of config file:

```
proxy_cache_path /var/cache/lemmy/voyager.lemmy.ml/ levels=1:2 keys_zone=lemmy_cache_voyager.lemmy.ml:10m max_size=100m use_temp_path=off;
```

And this next to `proxy_pass`:
```
proxy_cache lemmy_cache_voyager.lemmy.ml;
proxy_cache_use_stale error timeout updating http_500 http_502 http_503 http_504;
proxy_no_cache $cookie_jwt $cookie_auth $http_authorization;
proxy_cache_bypass $cookie_jwt $cookie_auth $http_authorization;
# for debugging, should probably be disabled in prod
add_header x-cache-status $upstream_cache_status;
```

Caching is enabled on https://voyager.lemmy.ml/ so you can give it a try and look at x-cache-status header.

ticoombs:
I've had a brief look at this previously, and had a look today. Looks like it works well. I had a MISS on the initial /communities and then a HIT immediately after. 
I also saw code in `lemmy` that changed the auth related $jwt and whether we sent the cache status or not. So I'm not entirely sure I would want to enable this cache just yet. 

I've enabled caching multiple times in $jobs and there are so many edge cases that it's sometimes difficult to find them without very thorough testing. (Mind you, I wasn't able to ask the dev's to change/fix bad items/code so I'm sure we can get everything sorted).

If we want to do this lets try a 1.4.0 release, after the dust settles on Lemmy 0.19.0.

## Tasks/Thoughts
- Ideally we would want to enable the cache in the nginx on the server level as that would reduce the amount of network traffic that the docker containers need to parse. That would unfortunately  mean we force people to use nginx as a proxy. Which I think is probably okay for lemmy-ansible, but might be an issue for others.
- How will the caching affect the other containers? (pictrs?)
- can we safely test PII/auth flows across multiple accounts
- if users have a CDN/proxy infront (ddos guard/etc) and all the CDN users come from X number of IPs how does this work then?

Nutomic:
[...]

Issue #32, opened 2022-03-28
Better way of avoiding port conflicts


rysiekpl:
*Continuing discussion from #3.*

It would be good to have a better way of avoiding port conflicts when running multiple instances of `lemmy`. Current approach (randomizing ports on every deploy) has [several downsides](#3).

Quick brainstorm about options here:

1. include `nginx` in the `docker-compose.yml`, allowing us to `proxy_pass` directly to specific containers; name relevant containers using the `{{domain}}`: `lemmy-{{domain}}`, `lemmy-ui-{{domain}}`, `pictrs-{{domain}}`, etc. Host-installed `nginx` would be unnecessary, or if already present would only need to hit a single port: the one exposed by the `nginx` container.

2. Have some faith in the admins running instances, and make it possible to explicitly set ports per deployment, such that and admin deploying 3 different Lemmy instances would just explicitly define three sets of ports. A version of this would be to have a "starting port" configurable per instance, and each actual service port just offset by a well-defined value (say, `lemmy_port` would be `starting_port`; `lemmy_ui_port` would be `starting_port` + 1, etc).

Option 1. is the cleanest, keeps all Lemmy deploys in a "package" within the `docker-compose.yml`, all managed from a single `lemmy-ansible` checkout with minimal side-effects on the host system. From a perspective of a sysadmin this might be the most preferable, and also is the most in-line with the "docker way of doing things" so to speak.

dessalines:
I definitely prefer option 2 for performance reasons... why have multiple nginx layers when you only need one.

rysiekpl:
> I definitely prefer option 2 for performance reasons... why have multiple nginx layers when you only need one.

Option 1. could/would be used with a single `nginx` layer, the difference would be that the `nginx` would be in the `docker-compose.yml`. The only reason for additional `nginx` on the host system would be if the system administrator wants to use it this way for whatever reason (other services running on the system, etc).

The important benefit of Option 1. is encapsulation. All services needed to run Lemmy are in that case encapsulated in and managed by `docker-compose.yml`, there are no side-effects on the host system. 

Performance hit should be negligible. I've run infrastructures with 2 `nginx` layers, serving hundreds or thousands of requests per second; `nginx` was never the bottleneck.

Nutomic:
We have an internal repo that we use to deploy lemmy.ml and associated test instances, it also uses the method 2 you explain. Just specify the starting port for each instance, and other ports are immediately above it. 

[...]

PR #237, opened 2024-04-13
feat: v1.5.0

ticoombs:
- postgres v16 (fixes: #227)
- pictrs 0.5.x
- nginx cache (fixes: #195)
- autoexplain & stats (fixes: #211) 
- lemmy_web_port (fixes: #32)


README.md
```diff
@@ -92,104 +92,8 @@ If you wish to see another distribution on the list, please test on the latest c
...
-- When upgrading from older versions of these playbooks, you will need to do the following:
-  - Rename `inventory/host_vars/{{ domain }}/passwords/postgres` file to `inventory/host_vars/{{ domain }}/passwords/postgres.psk`
-  - Copy the `examples/vars.yml` file to `inventory/host_vars/{{ domain }}/vars.yml`
-  - Edit your variables as desired
-- Run your regular deployment. Example: `ansible-playbook -i inventory/hosts lemmy.yml --become`
-
-### Upgrading to 1.1.0 (Lemmy 0.18.3)
-
-- No major changes should be required
+Since version `1.5.0` we no longer pollute the readme with upgrading instructions.
```
dessalines, line 95: Thx, I like this.

examples/config.hjson
```diff
@@ -3,8 +3,8 @@
   # https://join-lemmy.org/docs/en/administration/configuration.html
 
   database: {
-    host: postgres
-    password: "{{ postgres_password }}"
+    uri: "postgresql://lemmy:{{ postgres_password }}@postgres/lemmy"
```
dessalines, line 6: Maybe add this one as a comment, bc the original `host: postgres` should be good.

examples/config.hjson
```diff
@@ -3,8 +3,8 @@
...
+    pool_size: "10"
```
dessalines, line 7: Pry should be removed, and just use lemmy's default.

examples/vars.yml
```diff
@@ -35,6 +35,8 @@ postgres_env_vars:
   - POSTGRES_PASSWORD: "{{ postgres_password }}"
   - POSTGRES_DB: lemmy
 
+postgres_shm: 1g
```
dessalines, line 38: Maybe make this 2g then, to match above.

examples/hosts
```diff
@@ -10,7 +10,7 @@
 # - `lemmy_version`: <Optional> The back end version.
 # - `lemmy_ui_version`: <Optional> overrides the front end version.
 # - `pictrs_safety`: <Optional> If true, a docker container for pictrs-safety will be deployed and pict-rs will be configured to validate images through it. You will also need to set up a fedi-safety worker to validate the images.
-example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy pictrs_safety=false
+example.com  ansible_user=root domain=example.com  letsencrypt_contact_email=your@email.com  lemmy_base_dir=/srv/lemmy pictrs_safety=false lemmy_web_port=9633
```
dessalines, line 13: What's lemmy_web_port about? it'd need to be documented if you add that.

dessalines:
Also ansible format is finding some mis-indent.

lemmy.yml
```diff
@@ -241,15 +243,8 @@
...
-    #     chdir: /opt                    # changes to /opt directory
-    #     register: shell_output
-    # - debug: var=shell_output
-
     - name: Start docker-compose
-      community.docker.docker_compose:
+      community.docker.docker_compose_v2:
         project_src: "{{ lemmy_base_dir }}/{{ domain }}"
         state: present
         pull: true
```
jlandahl, line 250: As noted in https://github.com/LemmyNet/lemmy-ansible/pull/240 the `pull` parameter is no longer a boolean, so this value has to change from `true` to something like `always` or `missing` (it fails if the value is `true`). I don't know if this was a recent change on the Ansible side, but it tripped me up a couple of days ago.

templates/nginx.conf
```diff
@@ -60,6 +61,13 @@ server {
       proxy_set_header X-Real-IP $remote_addr;
       proxy_set_header Host $host;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+
+      proxy_cache lemmy_cache_{{domain|replace(".","_")|replace("-","_")}};
```
Nothing4You, line 65: this probably needs to ensure that at least the accept header is included in the [`proxy_cache_key`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_key) to avoid serving `activity+json` or html to the wrong client because it's already cached

dessalines:
We're shooting for the `0.19.4` release tomorrow, so this will also need to be merged with it, as it contains the upgrade instructions.

ticoombs:
@dessalines we're good to go

dessalines (approved):
Sweet, thx!

PR #238, opened 2024-04-21
Add cron to dependencies

marekm4:
Cron seems to be required:
```
TASK [Certbot renewal cronjob] *************************************************
fatal: [foresight.wiki]: FAILED! => {"changed": false, "msg": "Failed to find required executable \"crontab\" in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}
```

### DIFF
--- a/lemmy-almalinux.yml
+++ b/lemmy-almalinux.yml
@@ -96,6 +96,7 @@
         state: present
         name:
           - certbot
+          - cronie
           - curl
           - nginx
           - podman
--- a/lemmy.yml
+++ b/lemmy.yml
@@ -91,6 +91,7 @@
           - "python3-pip"
           - "virtualenv"
           - "python3-setuptools"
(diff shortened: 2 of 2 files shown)


codyro (changes requested):
I hadn't thought about `minimal` images not having this installed.

Would you adding it to the `lemmy-almalinux.yml` playbook as well so they're 1:1?

You would need to add `cronie` to the dependency list [here](https://github.com/LemmyNet/lemmy-ansible/blob/main/lemmy-almalinux.yml#L94-L111).

marekm4:
I was surprised with lack of cron as well

ticoombs:
Cron has only recently not been a default on debian iirc.

codyro (approved):

dessalines (approved):

PR #248, opened 2024-07-01
Fix prefix in nginx set_real_ip_from to only use private address range

Nothing4You:


Nothing4You:
If all private IPv4 RFC 1918 private IPs should be accepted then this could be extended to also take `10.0.0.0/8` and `192.168.0.0/16`.

codyro:
> If all private IPv4 RFC 1918 private IPs should be accepted then this could be extended to also take `10.0.0.0/8` and `192.168.0.0/16`.

I believe Docker uses the `172.17.0.0/16` by default, which might explain why this was configured this way. 

Podman uses `10.88.0.0/16` by default, so we should at least add that range as well, although I can see the argument for doing the full `/8`.

templates/nginx_internal.conf
```diff
@@ -10,7 +10,7 @@ http {
     resolver {{ nginx_internal_resolver }} valid=5s;
     # set the real_ip when from docker internal ranges. Ensuring our internal nginx
     # container can always see the correct ips in the logs
-    set_real_ip_from 172.0.0.0/8;
+    set_real_ip_from 172.16.0.0/12;
```
codyro, line 13: Please add another line for `10.88.0.0/16`, which Podman uses by default. Alternatively, add the default ranges as noted in your comment.

codyro (changes requested):

templates/nginx_internal.conf
Nothing4You: updated to all rfc 1918 private ranges now

dessalines (approved):

codyro (approved):

PR #249, opened 2024-07-03
feat: remove open port & enable docker live-restore

ticoombs:
Here we give a QOL update to lemmy-ansible.

- move to using a dedicated container by default for schedule, federate, and api
- close the open port that everyone seems to forgot about
- enable the live-restore so everyone can update docker without fear of their lemmy server going down
- cleanup a vim issue

### DIFF
--- a/templates/docker-compose.yml
+++ b/templates/docker-compose.yml
@@ -36,6 +36,53 @@ services:
 {% endif %}
     volumes:
       - ./lemmy.hjson:/config/config.hjson:Z
+    command:
+      - "--disable-scheduled-tasks"
+      - "--disable-activity-sending"
+    depends_on:
+      - postgres
+      - pictrs
+
+  lemmy-schedule:
+    image: {{ lemmy_docker_image }}
+    hostname: lemmy-schedule
+    restart: always
+    logging: *default-logging
+    environment:
+{% if lemmy_env_vars is defined and lemmy_env_vars|length > 0 %}
+{% for item in lemmy_env_vars %}
+{% for key, value in item.items() %}
+      - {{ key }}={{ value }}
+{% endfor %}
+{% endfor %}
+{% endif %}
+    volumes:
+      - ./lemmy.hjson:/config/config.hjson:Z
+    command:
+      - "--disable-activity-sending"
+      - "--disable-http-server"
+    depends_on:
+      - postgres
+      - pictrs
+
+  lemmy-federate:
+    image: {{ lemmy_docker_image }}
+    hostname: lemmy-federate
+    restart: always
+    logging: *default-logging
+    environment:
+{% if lemmy_env_vars is defined and lemmy_env_vars|length > 0 %}
+{% for item in lemmy_env_vars %}
+{% for key, value in item.items() %}
+      - {{ key }}={{ value }}
+{% endfor %}
+{% endfor %}
+{% endif %}
+    volumes:
+      - ./lemmy.hjson:/config/config.hjson:Z
+    command:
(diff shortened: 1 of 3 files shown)


Nothing4You:
It might make sense to put live restore behind a var that is disabled by default and instruct people to read and understand https://docs.docker.com/config/containers/live-restore/ first if they want to use it, otherwise it might be an easy trap for people with less docker experience.

> Live restore allows you to keep containers running across Docker daemon updates, but is only supported when installing patch releases (`YY.MM.x`), not for major (`YY.MM`) daemon upgrades.
>
> If you skip releases during an upgrade, the daemon may not restore its connection to the containers. If the daemon can't restore the connection, it can't manage the running containers and you must stop them manually.

templates/docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
       # actual and only port facing any connection from outside
       # Note, change the left number if port 1236 is already in use on your system
       # You could use port 80 if you won't use a reverse proxy
-      - "{{ lemmy_port }}:8536"
+      - "127.0.0.1:{{ lemmy_port }}:8536"
```
Nutomic, line 14: The comment needs updating then.

templates/docker-compose.yml
```diff
@@ -36,6 +36,53 @@ services:
...
+{% for key, value in item.items() %}
+      - {{ key }}={{ value }}
+{% endfor %}
+{% endfor %}
+{% endif %}
+    volumes:
+      - ./lemmy.hjson:/config/config.hjson:Z
+    command:
+      - "--disable-scheduled-tasks"
+      - "--disable-http-server"
```
dessalines, line 85: I could go either way on separating these... mostly I'm worried about the increased memory requirements on smaller servers, from running these 3 alongside each other.

Nutomic: Would be worth testing if it really makes much difference. Most likely overall memory usage would be similar, as you have the same data stored in memory either in a single process or split across multiple processes. But seeing as the only benefit here is to have separate logs, I would keep it in a single process. That way the docker-compose file is also easier to edit manually (its very easy to forget changing the version in one place).

dessalines: I agree then. 

@ticoombs if you want to show an example of the 3 servers just for documentation/examples sakes, you could move them into a differently named but unused `docker-compose.split_server.yml` or something, but leave this change out of the original file.

Nothing4You: if versions/image name is the concern this is easily solvable by yaml references, similar to what is already used e.g. in the lemmy woodpecker config: https://github.com/LemmyNet/lemmy/blob/f229f09f92f44bb67fce3801be5345d89a0b808a/.woodpecker.yml#L5

in a compose file you can specify a section with an `x-` prefix for this purpose: https://docs.docker.com/compose/compose-file/11-extension/

if resource usage is a concern it could be gated by a variable that is defaulting to false.

it would likely be beneficial to maintain a dedicated non-jinja compose file as reference for non-ansible-users, as people have already repeatedly been struggling with trying to understand the jinja templated file in here when they were trying to do non-ansible setups.

Nutomic: My main concern is about complexity, Id prefer to keep it simple.

> it would likely be beneficial to maintain a dedicated non-jinja compose file as reference for non-ansible-users, as people have already repeatedly been struggling with trying to understand the jinja templated file in here when they were trying to do non-ansible setups.

Yes definitely.

dessalines (approved):

dessalines (approved):

PR #256, opened 2025-01-22
Remove gzip compression from nginx config

Nutomic:
This is already handled in lemmy backend using [actix-web compression middleware](https://docs.rs/actix-web/latest/actix_web/middleware/struct.Compress.html) (also in 0.19). It also supports brotli and zstd besides gzip.

```bash
curl "https://lemmy.ml/api/v3/comment/list?post_id=3192572" -so /dev/null --compressed -v 2>&1 | grep content-encoding:
< content-encoding: br
```

### DIFF
--- a/templates/nginx.conf
+++ b/templates/nginx.conf
@@ -40,13 +40,6 @@ server {
     # Upload limit, relevant for pictrs
     client_max_body_size 20M;
 
-    # Enable compression for JS/CSS/HTML bundle, for improved client load times.
-    # It might be nice to compress JSON, but leaving that out to protect against potential
-    # compression+encryption information leak attacks like BREACH.
-    gzip on;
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Seems fine, although leaving it in is probably harmless too.

ticoombs:
We only gzip css/js/and svgs under this config.

Your solution:
Screenshot 2025-01-28 at 13 33 55

Current solution:
Screenshot 2025-01-28 at 13 44 18

From my understanding (and your POC) the actix-web does not handle the css/js files. As such using it here does not interfere with actix, but actually helps transfer less data overall. 

See client.js gets compressed from ~1850KB to 570KB! nearly a 70% reduction! Unless I'm mistaken.

Nutomic:
Youre right I didnt think of lemmy-ui, closing it then.

PR #260, opened 2025-03-19
0.19.10 release.

dessalines:


UPGRADING.md
```diff
@@ -6,6 +6,13 @@ While we specify a version of Lemmy, pict-rs, postgres, etc. at the point in tim
 
 While you are not forced into running the specific versions, we do not go through thorough testing on all version compatibility matrices, so please make your best judgement and always backup before performing updates.
 
+### Upgrading to 1.5.6 (Lemmy 0.19.10)
+
+#### Steps
+
+- `git pull && git checkout 1.5.6`
+- Run your regular deployment. Example: `ansible-playbook -i inventory/hosts lemmy.yml --become`
+
```
Nutomic, line 15: I wouldnt bother copy-pasting the same upgrade instructions for every version. Its not helpful and makes it harder to find the versions with manual steps required.

dessalines: It's necessary because of the release tag.

Nutomic: You mean the `git checkout *version*`? It would be enough to put that at the top as a general instruction, without repeating it every time.

dessalines: I spose we could, but many tagged versions will have to have their own specific upgrade instructions anyway.

ticoombs (approved):
I agree it's not great. Maybe we can move the discussion to an issue and decided what we want to do with v1. As going forward it would be better if we were consistent. Regardless of what it ends up looking like.

ticoombs:
I've also rolled this out with no issues.

PR #261, opened 2025-03-20
Avoid copy-pasted instructions in UPGRADING.md

Nutomic:
As discussed in https://github.com/LemmyNet/lemmy-ansible/pull/260

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -6,47 +6,11 @@ While we specify a version of Lemmy, pict-rs, postgres, etc. at the point in tim
 
 While you are not forced into running the specific versions, we do not go through thorough testing on all version compatibility matrices, so please make your best judgement and always backup before performing updates.
 
-### Upgrading to 1.5.6 (Lemmy 0.19.10)
-
-#### Steps
-
-- `git pull && git checkout 1.5.6`
-- Run your regular deployment. Example: `ansible-playbook -i inventory/hosts lemmy.yml --become`
-
-### Upgrading to 1.5.5 (Lemmy 0.19.9)
-
-#### Steps
-
-- `git pull && git checkout 1.5.5`
-- Run your regular deployment. Example: `ansible-playbook -i inventory/hosts lemmy.yml --become`
-
-### Upgrading to 1.5.4 (Lemmy 0.19.8)
-
-#### Steps
-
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -6,47 +6,12 @@ While we specify a version of Lemmy, pict-rs, postgres, etc. at the point in tim
 
 While you are not forced into running the specific versions, we do not go through thorough testing on all version compatibility matrices, so please make your best judgement and always backup before performing updates.
 
-### Upgrading to 1.5.6 (Lemmy 0.19.10)
+For each new version run the following steps in your local `lemmy-ansible` folder:
```
ticoombs, line 9: Lets add a heading above this, like: `## General Upgrade Instructions`

and maybe a little paragraph saying to double check the Specific Version instructions below, and if the version is not listed then it's 100% safe to deploy.

Something I also think we are missing is if there are database migrations in the version or not. As they would be the things an admin would want to lookout for. As these are the "Main" upgrade docs that Lemmy links too.

UPGRADING.md
```diff
@@ -6,47 +6,12 @@ While we specify a version of Lemmy, pict-rs, postgres, etc. at the point in tim
...
-#### Steps
-
-- `git pull && git checkout 1.5.1`
-- Run your regular deployment. Example: `ansible-playbook -i inventory/hosts lemmy.yml --become`
+- Find the [latest version](https://github.com/LemmyNet/lemmy-ansible/releases)
+- Fetch the updated files: `git pull && git checkout *version_name*`
+- Check below for version specific upgrade steps
+- Run the deployment: `ansible-playbook -i inventory/hosts lemmy.yml --become`
 
 ### Upgrading to 1.5.0 (Lemmy 0.19.4, Pict-rs 0.5, postgres 16)
```
ticoombs, line 16: Then another heading above here `## Specific Version Upgrades`

ticoombs (commented):
I think adding some extra headings will help lead people to the right sections.

UPGRADING.md
Nutomic: Updated

ticoombs:
Clean!

ticoombs (approved):

PR #268, opened 2025-08-21
update docker-compose template environment tag logic

bdonvr:
There is logic to add environment variables to the docker-compose.yml but an `environment:` tag is present regardless of whether there is any environment variables. Docker-Compose will throw an error for an empty `environment:` tag. (ex: `validating /srv/lemmy/lemmytest.local/docker-compose.yml: services.pictrs.environment must be a mapping`). This is especially an issue with pict-rs as a stock-deploy does not define any pict-rs variables and so docker will refuse to start the server.

 This commit moves the tag into the logic.

### DIFF
--- a/templates/docker-compose.yml
+++ b/templates/docker-compose.yml
@@ -24,8 +24,8 @@ services:
     hostname: lemmy
     restart: always
     logging: *default-logging
-    environment:
 {% if lemmy_env_vars is defined and lemmy_env_vars|length > 0 %}
+    environment:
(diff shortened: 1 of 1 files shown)


ticoombs:
This looks good from a code standpoint, but what item did you come across which caused this functionality?

Those values should (I thought) always be set. So it's interesting how you got a system that didn't need environment vars. Especially when they are basically required by postgres to even work.

codyro (approved):

PR #285, opened 2026-04-06
Adding an optional anubis config

dessalines:
This adds an optional [anubis](https://anubis.techaro.lol), which stands in front of lemmy-ui only. 

It doesn't block or impede federation, or apps, only the lemmy-ui front end.

I tested this via a fresh server, and it worked.


templates/botPolicy.yaml
```diff
@@ -0,0 +1,236 @@
...
+  ## If your backend service is running on the same operating system as Anubis,
+  ## you can uncomment this rule to make the challenge easier when the system is
+  ## under low load.
+  ##
+  ## If it is not, remove weight.
+  # - name: low-load-average
+  #   action: WEIGH
+  #   expression: load_15m <= 4.0 # make sure to end the load comparison in a .0
+  #   weight:
+  #     adjust: -10
```
Nutomic: These are probably not going to work in Docker so you can remove them. You can also remove most other things which are commented out, and link to the official documentation instead.

dessalines: @povoq you use anubis on a prod server right? Is there any "tweaked for lemmy" anubis config that people have come up with?

dessalines:
I tested this via a fresh server, and it works.

I'd like someone who knows their way around anubis, and hopefully who's discovered a better "lemmy-specific" anubis config to tweak our `botPolicy.yaml`, before messing with it too much.

templates/botPolicy.yaml
dessalines: I asked [over here](https://lemmy.ml/post/45684888), and it doesn't seem like anyone came up with anything specific. I'll remove the commented lines then at least.

dessalines:
@ticoombs @codyro also let me know if this looks good.

lemmy.yml
```diff
@@ -269,6 +269,9 @@
             - src: "templates/nginx.conf"
               dest: "/etc/nginx/sites-available/{{ domain }}.conf"
               mode: "0644"
+            - src: "templates/botPolicy.yaml"
+              dest: "{{ lemmy_base_dir }}/{{ domain }}/botPolicy.yaml"
+              mode: "0644"
```
ticoombs, line 280: I think this needs to be moved to its own section and have it's own notify, as currently it wont reload anubis if the file already is there and its updated. 

(unless anubis watches the configuration and auto-reloads, but i didnt see that in the docs)

templates/botPolicy.yaml
```diff
@@ -0,0 +1,58 @@
...
+  considerHost: false
+  ttl: 24h
+
+status_codes:
+  CHALLENGE: 200
+  DENY: 200
+
+store:
+  backend: memory
+  parameters: {}
```
ticoombs, line 208: how big can the memory grow? Annnd I answered my own question.

> https://anubis.techaro.lol/docs/admin/policies/#memory
> The in-memory backend exists mostly for validation, testing, and to ensure that the default configuration of Anubis works as expected. Do not use this persistently in production.

Out of all the options, besides from valkey, the best would be bbolt i think (filesystem db). As we are suggesting the best option for single-host usecases. (And the s3 option could end up hurting people's wallets with the amount of api requests)

```suggestion
store:
  backend: bbolt
  parameters:
    path: /data/anubis.bdb
```

templates/docker-compose.yml
```diff
@@ -136,3 +140,26 @@ services:
...
+    healthcheck:
+      test: ["CMD", "anubis", "--healthcheck"]
+      interval: 5s
+      timeout: 30s
+      retries: 5
+      start_period: 500ms
+    ports:
+      - 8924:8924
+    volumes:
+      - "./botPolicy.yaml:/data/cfg/botPolicy.yaml:ro"
```
ticoombs, line 164:
```suggestion
      - "./botPolicy.yaml:/data/cfg/botPolicy.yaml:ro"
      - "./anubis.bdb:/data/anubis.bdb:rw"
      
```

lemmy.yml
ticoombs, line 289:
```suggestion
        - name: Add anubis bot policy
          ansible.builtin.template:
            src: "templates/botPolicy.yaml"
            dest: "{{ lemmy_base_dir }}/{{ domain }}/botPolicy.yaml"
            mode: "0644"
          notify: reload anubis
```

ie move it here

ticoombs: and then add a notify file for `docker compose restart anubis`

ticoombs (changes requested):
Some small changes needed. Otherwise it looks good.

Nutomic:
Bump, seems like this still needs some changes.

templates/botPolicy.yaml
nrekoo: 1 datapoint: on a moderately high traffic instance with constrained disk I/O (ie, HDD), bbolt was not nearly performant enough and resulted in anubis slowing to a crawl, with challenges failing after the 504 timeout was reached by nginx. 

I don't know if this would be the case on small personal instances. I think it could be, but only in niche cases. where production-ready in-memory db is needed, anubis does support valkey/redis.

templates/docker-compose.yml
nrekoo: This config does not work out of the box for at least two reasons. 
- If anubis.bdb does not already exist somehow it gets created and mounted as a directory not a file
- If anubis.bdb does exist, it is mounted in but owned by root.

Both result in this error:
```
anubis-1  | 2026/05/21 16:58:47 can't parse policy file: can't parse policy file /data/cfg/botPolicy.yaml: config is not valid:
anubis-1  | store: configuration is invalid: bbolt: can't write to path
```

I ended up mounting a directory and creating anubis.bdb inside it, owned by uid/gid 1000. There may be a cleaner solution though.

dessalines:
Sry, I'll try to get to this soon.

lemmy.yml
dessalines: Any updates to these files seem to restart all the docker services.

templates/botPolicy.yaml
dessalines: Good call, I'll try this out.

> bbolt was not nearly performant enough and resulted in anubis slowing to a crawl, with challenges failing after the 504 timeout was reached by nginx.

If bbolt isn't performant enough despite being their main recommendation, then that should be opened up as an issue on the anubis repo.

lemmy.yml
dessalines: K I'll try it out.

templates/docker-compose.yml
dessalines: I just tried for like half an hour, and could never get past this error. Tried different perms, manually creating the file, nothing worked.

I've pushed up my changes, but it looks like bbolt doesn't work with docker.

dessalines:
I've pushed up my changes, but I couldn't get bbolt working with docker. Always gives this error:

```
anubis-1  | can't parse policy file: can't parse policy file /data/cfg/botPolicy.yaml: config is not valid:
anubis-1  | store: configuration is invalid: bbolt: can't write to path
```

Even though the docs say that anubis's default in-memory DB shouldn't be used in production, we've been using it fine on prod instances for over a month now, so I think its fine.

If someone else wants to take a stab at getting bbolt working, they can, but otherwise I think we should just merge with the in-memory version.

templates/botPolicy.yaml
dessalines: Nevermind, couldn't get bbolt working with docker.

templates/docker-compose.yml
```diff
@@ -136,3 +140,28 @@ services:
...
+      OG_EXPIRY_TIME: "24h"
+      ED25519_PRIVATE_KEY_HEX: "d1c09841dc1baff09ddce251bce31e89d7b8b8d055a2d87bba1637ebac3dcc62"
+    healthcheck:
+      test: ["CMD", "anubis", "--healthcheck"]
+      interval: 5s
+      timeout: 30s
+      retries: 5
+      start_period: 500ms
+    ports:
+      - 8924:8924
```
pe1uca, line 162: I don't think exporting anubis's port is necessary.  
The requests are still handled by nginx and forwarded to anubis.  

My personal instance has been working properly without exposing the ports.

dessalines: I think its necessary because the `nginx_internal.conf` still needs to have access to it. IE `proxy_pass "http://anubis:8924";`

dessalines:
@ticoombs I tried getting anubis to work with bbolt, but failed as you can see above. 

Since there are some large instances already working with its default memory version, I think I'd like to just use that, as I can verify that it works.

ticoombs (approved):

Nutomic:
I'm starting to wonder if Anubis is really worth using anymore, because it seems that web crawler developers have long adapted. It is really trivial to detect that a site uses Anubis, and then have a powerful server solve the challenge. At the same time it adds a significant delay for real humans. Recently I keep getting Anubis challenges with difficulty 6 which take more than a minute to complete on a ~5 year old laptop.

In case of lemmy.ml we also added Anubis to reduce server load. But after some users complained about it, we removed Anubis and CPU usage remained completely unchanged. It seems that a few lines of nginx config to block different user-agents achieve the same effect, and users dont notice it at all:
```nginx
if ($http_user_agent = "") { return 444; }
if ($http_user_agent = "-") { return 444; }
if ($http_user_agent = "matrix-media-repo") { return 444; }
if ($http_user_agent ~* (dotbot|ahrefsbot|dataforseo-bot|amazonbot|semrushbot|yandexbot|bingbot|petalbo>
return 444;
}
```

Here is also a blog post about this topic: https://fzakaria.com/2026/07/09/who-does-anubis-actually-stop

dessalines:
> Recently I keep getting Anubis challenges with difficulty 6 which take more than a minute to complete on a ~5 year old laptop.

We'd have to see if we keep getting those on [newer versions of anubis](https://github.com/TecharoHQ/anubis/releases), and if not, see if we've misconfigured something, or open up an issue on anubis to see why we're getting high difficulties by default.

> It seems that a few lines of nginx config to block different user-agents achieve the same effect, and users dont notice it at all:

I'm not against adding these, but they should be a separate PR. This PR only adds anubis as an option, which is off by default

> I'm starting to wonder if Anubis is really worth using anymore, because it seems that web crawler developers have long adapted. It is really trivial to detect that a site uses Anubis, and then have a powerful server solve the challenge.

I read that post, and it relies on their crawler solving then storing keys (which shouldn't last more than a few days anyway) for every site visited. I'm not convinced that forcing crawlers to pay a CPU tax isn't effective, and I trust the many people collaborating on these anti-crawler tools to run these crawlers / testing suites against their own projects to test its effectiveness. Most importantly, I don't want to try to solve this problem ourselves, I want to rely on a project or tool dedicated solely to this problem.

dessalines:
I finally figured out the reason anubis was using difficulty `6` . 

On some production instances, we're using this in `nginx_internal.conf`:

`proxy_set_header X-Real-IP $remote_addr;`

This is incorrect, as that `remote_addr` ends up being the docker internal IP. Doing some research, the standard now seems to be to never set the Real-IP when using docker, and only using the `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;`

If incorrectly setting `X-Real-IP`, anubis ends up using an internal docker IP:

```
{"function":"github.com/TecharoHQ/anubis/lib.(*Server).checkRules","file":"github.com/TecharoHQ/anubis/lib/anubis.go","line":164},"msg":"explicit deny","user_agent":"Opera/9.47.(X11; Linux x86_64; quz-PE) Presto/2.9.173 Version/10.00","accept_language":"en","priority":"","x-forwarded-for":"CORRECT","x-real-ip":"INTERNAL_DOCKER_IP","check_result":{"name":"bot/deny-aggressive-brazilian-scrapers","rule":"DENY"}}
```

Just removing that line above made anubis use the x-forwarded-for IP for both in its internal logging lines, and now I get difficulty = 2 always.

We already have `X-Real-IP` removed in lemmy-ansible, so it shouldn't be a problem.

PR #289, opened 2026-05-05
Fetch latest Lemmy version over HTTP, delete VERSION file

Nutomic:
This way we dont need to make a new release in this repo for every Lemmy release.


dessalines:
We should def keep this hardcoded and do tagged releases, because they could include specific instructions (especially for things like config changes, etc), and each release should be pinned to a specific version, so they can checkout a specific tag and know its going to be that lemmy version.

Nutomic:
Makes sense for major versions. For minor versions we could upgrade it automatically in this way, but that would get more complicated.

ticoombs:
A release is a point-in-time snapshot. Moving to this system will remove that constraint and cause versions moving forward to potentially be out of sync like dess suggested. 

The idea has merit though, I wonder if instead we can link it to a 1.0.x version? Then it would allow patches to be released by lemmy similar to how we allow pictrs to have any 0.5.x version. As unless someone completely re-writes this ansible, I foresee general changes to lemmy-ansible would be minor and general upkeep moving forward.

Nutomic:
This PR is currently using the [latest release](https://docs.github.com/en/rest/releases/releases?apiVersion=2026-03-10#get-the-latest-release) endpoint, which only returns the single last release. This could be a problem if we ever publish 0.19.x version after 1.0 is out. So to make it work we would have to use the [list releases](https://docs.github.com/en/rest/releases/releases?apiVersion=2026-03-10#list-releases) endpoint and pick the right version from there, or never publish another 0.19 version after 1.0 is out (and the same problem would happen once we publish a hypothetical stable 2.0 version).

dessalines:
Doing tagged lemmy-ansible releases doesn't take too long. It could even be added as part of the lemmy `release.bash` script if the time spent is what you're worried about.

But ya I think that releases should be snapshots in time, and not do http fetches to get versions.

PR #292, opened 2026-06-08
In nginx config use `$remote_addr`

Nutomic:


files/proxy_params
```diff
@@ -1,4 +1 @@
-proxy_set_header Host $http_host;
-proxy_set_header X-Real-IP $remote_addr;
-proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-proxy_set_header X-Forwarded-Proto $scheme;
\ No newline at end of file
+proxy_set_header X-Forwarded-For $remote_addr;
```
Nutomic, line 1: Lemmy doesnt use any of these other headers so it seems they can be safely removed.

templates/nginx.conf
```diff
@@ -56,9 +56,7 @@ server {
       proxy_http_version 1.1;
       proxy_set_header Upgrade $http_upgrade;
       proxy_set_header Connection "upgrade";
-      proxy_set_header X-Real-IP $remote_addr;
-      proxy_set_header Host $host;
```
dessalines, line 60: I wouldn't remove these, they might be necessary for apps, and it might cause problems.

Just the `forwarded-for` should be changed.

Nutomic: These headers are added by nginx to requests going to the Lemmy backend. They are not used when sending responses back to the client. 

I already changed it on lemmy.ml and commented these lines out. Rate limits are still working fine.

dessalines (approved):

PR #294, opened 2026-06-09
Add back Host header

Nutomic:
This is actually necessary, removing it broke incoming federation on lemmy.ml. Reverts part of https://github.com/LemmyNet/lemmy-ansible/pull/292.


Nutomic:
Merging this directly so other instances dont get broken. Do we need to make a new release? Seems like people just use main branch anyway.

PR #296, opened 2026-06-10
Make sure the nginx_internal conf uses proxy_add_x_forwarded_for

### LINKED ISSUE #295: proxy_params breaks rate limiting with double-nginx setup (host nginx + internal Docker proxy)
Hi, I just upgraded to 0.19.19 via ansible (tag 1.5.16) and the instance broke immediately after. All pages return 500 with rate_limit_error in lemmy-ui logs.

Root cause: the security fix changed proxy_params to use $remote_addr for X-Forwarded-For, but lemmy-ansible deploys two nginx layers (host + internal Docker proxy). The internal proxy's $remote_addr is the Docker gateway IP, not the user so Lemmy sees all traffic from one IP and saturates the rate limit instantly.

Workaround: change proxy_params on the server back to $proxy_add_x_forwarded_for and restart the proxy container.

dessalines:
- Fixes #295

### DIFF
--- a/files/proxy_params
+++ b/files/proxy_params
@@ -1,4 +1,3 @@
 proxy_set_header Host $http_host;
-proxy_set_header X-Real-IP $remote_addr;
+# Needs to be proxy_add_x_forwarded_for, not remote_addr, which is the docker IP.
 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-proxy_set_header X-Forwarded-Proto $scheme;
--- a/files/proxy_params
+++ b/files/proxy_params
@@ -1,2 +1,4 @@
-proxy_set_header Host $host;
-proxy_set_header X-Forwarded-For $remote_addr;
+proxy_set_header Host $http_host;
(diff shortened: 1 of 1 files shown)


files/proxy_params
```diff
@@ -1,2 +1,4 @@
-proxy_set_header Host $host;
-proxy_set_header X-Forwarded-For $remote_addr;
+proxy_set_header Host $http_host;
+proxy_set_header X-Real-IP $remote_addr;
+proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+proxy_set_header X-Forwarded-Proto $scheme;
```
Nutomic: Why are you adding these headers back? `X-Real-IP` and `X-Forwarded-Proto` arent used anywhere.

Nutomic: And add a comment why this one needs to use `$proxy_add_x_forwarded_for`.

dessalines: Alright, although I can't find any examples of proxy_params without these 4 lines.

Nutomic (approved):

